PR #43, opened 2020-02-28
Resolve application monitoring on Windows

### LINKED ISSUE #42: Applications not shown on Windows
On Windows, no applications are shown in the Dapr applications tree view even when verified running.  Also, the `daprd-down` task is not able to kill running Dapr applications.

philliphoff:
The `ps-list` library used to monitor running `daprd` processes does not provide the command line when used on Windows and, with other process-level differences between the platforms, we are unable to properly detect Dapr applications nor shut them down.  We now use the `wmic` command on Windows to pull that information.

Resolves #42.

### DIFF
--- a/src/services/processProvider.ts
+++ b/src/services/processProvider.ts
@@ -1,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as psList from 'ps-list';
+import * as os from 'os';
+import * as sysinfo from 'systeminformation';
+import { Process } from '../util/process';
+
+export interface ProcessInfo {
+    cmd: string;
+    name: string;
+    pid: number;
+}
+
+export interface ProcessProvider {
+    listProcesses(name: string): Promise<ProcessInfo[]>;
+}
+
+export class UnixProcessProvider implements ProcessProvider {
+    async listProcesses(name: string): Promise<ProcessInfo[]> {
+        const processes = await psList();
+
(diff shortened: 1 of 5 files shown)


src/services/processProvider.ts
```diff
@@ -0,0 +1,69 @@
...
+
+function getWmicValue(line: string): string {
+    const index = line.indexOf('=');
+
+    return line.substring(index + 1);
+}
+
+export class WindowsProcessProvider implements ProcessProvider {
+    async listProcesses(name: string): Promise<ProcessInfo[]> {
+        const list = await Process.exec(`wmic process where "name='${name}.exe'" get commandline,name,processid /format:list`);
```
bwateratmsft: Ugh, is WMI really the only way to get a process list?!

philliphoff: Ugh is correct.  It's not the only way to get a process list, but it *is* the only way I've found to get the command line associated with a given process.  The library I was using can get the command line for Unix-y platforms, but only PID for Windows.  All of the libraries I've seen that can get the command line for Windows appear to use `wmic` under the covers, but I didn't find any library that seemed to have a significant following.  Since our needs are fairly specific, I chose to use it directly.

Future versions of the Dapr CLI will allow us to run the `dapr` CLI instead of starting the `daprd` daemon directly, which means we'll be able to use a `dapr` CLI to retrieve the running application list, which will allow us to dispense with this process list logic altogether.

PR #64, opened 2020-03-11
Show error when redirected to HTTPS endpoint

### LINKED ISSUE #44: Make Dapr invocation work with .NET Core ASP.NET using SSL
Invocation of methods via Dapr fails with a certificate validation error when using Dapr with a .NET Core ASP.NET application configured for (forcing) SSL.  As the default for newly scaffolded .NET Core ASP.NET applications is to use SSL, we should see if we can make that "just work".

philliphoff:
Shows an error when the application (during a GET or POST method invocation) redirects to a local HTTPS endpoint (which is most likely to be a configuration error, as Dapr expects calls to be invoked via HTTP). Also now treats all other redirection as the "result" of the invocation rather than following the redirect and returning *that* result.

This addresses (for now) the issue where the users forgets the `--no-https` flag when scaffolding a new .NET Core web application, which otherwise scaffolds a blanket redirection of calls to the HTTP endpoint to the HTTPS endpoint. 

This resolves #44.

### DIFF
--- a/src/services/daprClient.ts
+++ b/src/services/daprClient.ts
@@ -4,13 +4,32 @@
 import * as url from 'url';
 import * as vscode from 'vscode';
 import { DaprApplication } from "./daprApplicationProvider";
-import { HttpClient } from './httpClient';
+import { HttpClient, HttpResponse } from './httpClient';
 import { localize } from '../util/localize';
 
 export interface DaprClient {
     invokeGet(application: DaprApplication, method: string, token?: vscode.CancellationToken): Promise<unknown>;
     invokePost(application: DaprApplication, method: string, payload?: unknown, token?: vscode.CancellationToken): Promise<unknown>;
     publishMessage(application: DaprApplication, topic: string, payload?: unknown, token?: vscode.CancellationToken): Promise<void>;
+}
+
+function manageResponse(response: HttpResponse): unknown {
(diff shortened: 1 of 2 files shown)


src/services/daprClient.ts
```diff
@@ -1,34 +1,55 @@
...
+function manageResponse(response: HttpResponse): unknown {
+    if (response.status >= 300 && response.status < 400) {
+        const redirectionUrl = response.headers['location'];
+
+        if (redirectionUrl) {
+            const parsedRedirectionUrl = new url.URL(redirectionUrl);
+
+            if (parsedRedirectionUrl.protocol === 'https:'
+                && (parsedRedirectionUrl.hostname === 'localhost' || parsedRedirectionUrl.hostname === '127.0.0.1')) {
+                throw new Error(localize('services.daprClient.httpsRedirectError', 'The application redirected to the local HTTPS endpoint \'{0}\'. This is usually a mistake in application configuration. Dapr expects to invoke the application via HTTP.', redirectionUrl));
```
bwateratmsft: Is Dapr really HTTPS-free?

philliphoff: The Dapr sidecar is intended to be co-located within the same network namespace (i.e. localhost) as the running application, which generally negates the need for HTTPS between it and the application.  (Communication between Dapr sidecars can optionally be secured, however.)

PR #92, opened 2020-04-24
Scaffold only folder-scoped tasks/configurations.

### LINKED ISSUE #88: dapr scaffold dapr tasks not working in workspace view
The command `Dapr: Scaffold Dapr Tasks` produces the error `Open a folder or workspace with a debug launch configuration.` in vscode's workspace view (after doing a `Open Workspace...`). The command works as expected in the folder view (after doing a `Open Folder...`) with the same folders that are in the workspace.

This is on vscode 1.44.2 with the 0.2.1 extension.

philliphoff:
Like [`vscode-docker`](https://github.com/Microsoft/vscode-docker), to avoid VS Code's quirks with respect to tasks/configurations, we should only scaffold folder-scoped tasks/configurations even when a workspace is opened.

Resolves #88.

### DIFF
--- a/src/commands/scaffoldDaprTasks.ts
+++ b/src/commands/scaffoldDaprTasks.ts
@@ -45,6 +45,7 @@ interface ScaffoldWizardContext {
     appId: string;
     appPort: number;
     configuration: vscode.DebugConfiguration;
+    folder: vscode.WorkspaceFolder;
 }
 
 const DefaultPort = 80;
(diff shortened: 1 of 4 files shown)


src/commands/scaffoldDaprTasks.ts
```diff
@@ -109,12 +110,20 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
 
     const configurationStep: WizardStep<ScaffoldWizardContext> =
         async wizardContext => {
-            const workspaceConfigurations = getWorkspaceConfigurations();
+            const folder = vscode.workspace.workspaceFolders?.[0];
```
bwateratmsft, line 113: If an array is defined but empty, does `arr[0]` throw? Or just return `undefined`?

philliphoff: In JS, `array[property]` for an undefined property returns `undefined`, which is handy for certain error scenarios and very different from .NET.

bwateratmsft: `ArgumentOutOfRangeException`!

PR #102, opened 2020-07-23
Add default for componentsPath property.

### LINKED ISSUE #98: daprd no longer defaults to a components path
With 0.9.0, `daprd` no longer has a default components path (previously `./components`).  See [dapr/dapr#1722](https://github.com/dapr/dapr/pull/1722).  This now means it's the responsibility of the tooling to set an appropriate `--components-path` argument.  The `daprd` task should be updated to set the argument the previously-default `daprd` value.  (Alternatively, the task scaffolding can scaffold the task `componentsPath` property.)

The workaround is to set the `componentsPath` property (e.g. to `./components`).

philliphoff:
Adds `./components` as a default for the `componentsPath` property of the `daprd` task, as the `daprd` CLI no longer uses that as a default, if omitted.

Resolves #98.

### DIFF
--- a/src/tasks/daprdCommandTaskProvider.ts
+++ b/src/tasks/daprdCommandTaskProvider.ts
@@ -59,7 +59,7 @@ export default class DaprdCommandTaskProvider extends CommandTaskProvider {
                                 .withNamedArg('--alsologtostderr', daprDefinition.alsoLogToStdErr, { assignValue: true })
                                 .withNamedArg('--app-id', daprDefinition.appId)
                                 .withNamedArg('--app-port', daprDefinition.appPort)
-                                .withNamedArg('--components-path', daprDefinition.componentsPath)
+                                .withNamedArg('--components-path', daprDefinition.componentsPath ?? './components')
                                 .withNamedArg('--config', daprDefinition.config)
                                 .withNamedArg('--control-plane-address', daprDefinition.controlPlaneAddress)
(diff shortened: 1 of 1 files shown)


src/tasks/daprdCommandTaskProvider.ts
```diff
@@ -59,7 +59,7 @@ export default class DaprdCommandTaskProvider extends CommandTaskProvider {
                                 .withNamedArg('--alsologtostderr', daprDefinition.alsoLogToStdErr, { assignValue: true })
                                 .withNamedArg('--app-id', daprDefinition.appId)
                                 .withNamedArg('--app-port', daprDefinition.appPort)
-                                .withNamedArg('--components-path', daprDefinition.componentsPath)
+                                .withNamedArg('--components-path', daprDefinition.componentsPath ?? './components')
```
bwateratmsft, line 62:
```suggestion
                                .withNamedArg('--components-path', daprDefinition.componentsPath || './components')
```
Unless `''` (empty string) is a valid value, I'd use `||` instead of `??`.

philliphoff: Since we're basically acting as a pass-through, I'm inclined to not do too much interpretation of the user's intent and provide a default only when the user actually omits a value, allowing the underlying CLI to determine what `''` ultimately represents.

PR #106, opened 2020-08-25
Update to get new serialize-javascript

bwateratmsft:
This update gets a newer version of serialize-javascript (among other things) to fix the security vulnerability.


philliphoff:
Looks like `vsce` doesn't like that you're using `vscode` types newer than the specified `vsce` engine.  Aside from that, looks ok.

bwateratmsft:
> Looks like `vsce` doesn't like that you're using `vscode` types newer than the specified `vsce` engine. Aside from that, looks ok.

Gah! Every time I do `npm update` I forget this.

PR #108, opened 2020-08-26
Resolve localization issues

### LINKED ISSUE #104: It shows "Failed to load message bundle..." after expanding "APPLICATIONS"
**OS:** Win10/Mac
**Build Version:** 20200725.1

**Repro Steps:**

1. F1 --> Extensions: Install from VSIX... --> Select .vsix file of Dapr.
2. Expand "APPLICATIONS" and check the result.

**Expect:**
It shows locally-running Dapr applications.

**Actual:**
It shows "Failed to load message bundle. See console for details".
image

**More Info:**
This issue does not reproduce when installing a release Dapr build on the same machine.

philliphoff:
Some update (or combination of updates) caused our integration of `vscode-nls` for localization to break.  I'm still at a loss to understand what, exactly, changed, but it did expose that we weren't necessarily using the `vscode-nls` library totally correctly.  Only after a lot of fiddling was I able to find a combination of build and runtime logic that seems to work for our configuration(s).  Specifically, switching from separate localization files to "metadata bundles" and then accommodating the way `vscode-nls` expects to locate those bundles even when we webpack our source for release.

Resolves #104.


bwateratmsft:
@philliphoff do you know to what degree this will affect Docker?

philliphoff:
> @bwateratmsft do you know to what degree this will affect Docker?

If and when we *actually* want to localize that extension, we'll need to do similar work as done here.  For example, the build tasks aren't generating the localization metadata at all, nor are the localization calls being rewritten (by `vscode-nls-dev`) to use the metadata.  Unfortunately, due to the way `vscode-nls-dev` works, you can't "hide" it behind an adaptor type as is being done today (in the Docker extension and other Azure extensions).  I think the only reason why the Dapr extension was affected was that I had taken localization setup a couple of steps further than our other extensions.

bwateratmsft:
@philliphoff Do you know if there are any consumers that are actually going all the way to proper localization with `vscode-nls`?

PR #118, opened 2020-09-01
Update per latest Dapr pub/sub spec.

### LINKED ISSUE #117: Pop up an error "Request failed with status code 404" when publishing message
**OS:** Windows
**Build Version:** 20200831.7

**Repro Steps:**
1. Create a new .NET Core 3.1 ASP.NET Web API application by "dotnet new webapi --no-https --name netwebapp".
2. Open the application in VS Code --> Select "Dapr: Scaffold Dapr Tasks" from Command Palette --> Select ".NET Core Launch (web)" when asked to select a configuration --> Enter an application ID --> Enter an application port.
3. Debug with ".NET Core Launch (web) with Dapr".
4. Create the file "Controllers/TopicController.cs" file and type the below content.
```
using Microsoft.AspNetCore.Mvc;
		
		namespace netwebapp.Controllers
		{
		    [ApiController]
		    public class TopicController : ControllerBase
		    {
		        [HttpGet("dapr/subscribe")]
		        public string[] GetDaprSubscriptions() {
		            return new[] { "topicA" };
		        }
		
		        [HttpPost("topicA")]
		        public ActionResult HandleTopicA([FromBody] object body) {
		            return this.Ok();
		        }
		    }
		}
```
5. Navigate to the Dapr Explorer --> Right click the application --> Select "Publish Message to Application" --> Enter the topic and a payload.

**Expect:**
Succeed to publish message to application.

**Actual:**
Pop up an error.
image

**More Info:**
This issue also reproduces for node.js project.

philliphoff:
At some point, the Dapr publish/subscribe spec changed to require specifying the "pubsubname" (i.e. the component used to perform the publish/subscribe), both when subscribing from the application code as well as when building the URL to publish a message.

This change updates the publish message command to prompt the user for the "pubsubname" and uses that in the URL used to publish the message.

Resolves #117.

### DIFF
--- a/src/commands/applications/publishMessage.ts
+++ b/src/commands/applications/publishMessage.ts
@@ -13,8 +13,30 @@ import { getLocalizationPathForFile } from '../../util/localization';
 
 const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
 
+const publishMessagePubSubNameStateKey = 'vscode-docker.state.publishMessage.pubSubName';
 const publishMessageTopicStateKey = 'vscode-docker.state.publishMessage.topic';
 const publishMessagePayloadStateKey = 'vscode-docker.state.publishMessage.payload';
+
+export async function getPubSubName(context: ITelemetryContext, ui: UserInput, workspaceState: vscode.Memento): Promise<string> {
+    const previousMethod = workspaceState.get<string>(publishMessagePubSubNameStateKey);
+
+    context.properties.cancelStep = 'pubSubName';
+
+    const topic = await ui.showInputBox(
+        {
+            prompt: localize('commands.publishMessage.pubSubNamePrompt', 'Enter the publish/subscribe component name used to publish'),
+            value: previousMethod,
+            validateInput: value => {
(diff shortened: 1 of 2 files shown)


src/commands/applications/publishMessage.ts
```diff
@@ -13,9 +13,31 @@ import { getLocalizationPathForFile } from '../../util/localization';
...
+    const previousMethod = workspaceState.get<string>(publishMessagePubSubNameStateKey);
+
+    context.properties.cancelStep = 'pubSubName';
+
+    const topic = await ui.showInputBox(
+        {
+            prompt: localize('commands.publishMessage.pubSubNamePrompt', 'Enter the publish/subscribe component name used to publish'),
+            value: previousMethod,
+            validateInput: value => {
+                return value === ''
```
bwateratmsft, line 30: Is `undefined` a valid value?

philliphoff: My understanding is that the wizard input method always returns a valid (if empty) string.

PR #132, opened 2020-12-15
Update task definitions for Dapr 1.0

### LINKED ISSUE #129: Match deprecation of `daprd` `--placement-address` property
The latest version of `daprd` deprecates the `--placement-address` property in favor of a new `--placement-host-address` property.  We need to update the extension to begin using this new property; it appears that the new 1.0.0 RC versions of Dapr drop the old property altogether.

philliphoff:
Updates the definitions for `dapr` and `daprd` tasks to match the exposed arguments of the 1.0 versions of the CLIs.  Splits the scaffolding of project-level Dapr components into a separate command as this is now handled at the user-level by the `dapr init` CLI command.  (I suspect there will be little need for such scaffolding in the future, but it was easy enough to keep around for now.)  Component scaffolding now includes the newer Zipkin component.

Resolves #129.


src/commands/scaffoldDaprComponents.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getLocalizationPathForFile } from '../util/localization';
+import { ConflictHandler } from '../scaffolding/conflicts';
+
+const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
+
+async function scaffoldDaprComponents(scaffolder: Scaffolder, templateScaffolder: TemplateScaffolder): Promise<void> {
+    const rootWorkspaceFolderPath = (vscode.workspace.workspaceFolders ?? [])[0]?.uri?.fsPath;
+
+    if (!rootWorkspaceFolderPath) {
+        throw new Error(localize('commands.scaffoldDaprTasks.noWorkspaceError', 'Open a folder or workspace.'));
```
karolz-ms, line 20: Just wondering if this is not too terse? Alternative would be something along the lines of "To create Dapr component files, you must open a folder or VS Code workspace first".

karolz-ms (approved):
Just one question to consider, otherwise looks great!

src/commands/scaffoldDaprComponents.ts
philliphoff: Sounds reasonable.  Updated.

PR #149, opened 2021-03-25
Add settings for explicit Dapr binary paths

### LINKED ISSUE #139: Allow users to explicitly set path to Dapr binaries
The installer for current versions of Dapr (including 1.0.0) no longer put its binaries in `PATH`; they ask users to do that themselves, if desired.  This means that the extension may not always be able to find it, and it's not always obvious to users *how* to fix it themselves.  (In particular, Mac is complicated because it offers multiple shells and `PATH` may be diffferent depending on how the terminal was created.)

The extension should offer a setting where the user can explicitly set the path to the Dapr binaries, as a workaround for when they cannot be found in the `PATH` (or, when, the user has explicitly chosen not to add them to the `PATH`).

philliphoff:
Adds configuration settings for users with Dapr binaries not in the `PATH`.  Currently, the extension expected `dapr` and `daprd` to be in the `PATH`.  This is not always the case (and, sometimes, is non-intuitive as to how to add them to the "right" `PATH`).  In that case, users can now just set an explicit path to the binary.

Resolves #139.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -29,6 +29,7 @@ import HandlebarsTemplateScaffolder from './scaffolding/templateScaffolder';
 import LocalScaffolder from './scaffolding/scaffolder';
 import NodeEnvironmentProvider from './services/environmentProvider';
 import createScaffoldDaprComponentsCommand from './commands/scaffoldDaprComponents';
+import VsCodeSettingsProvider from './services/settingsProvider';
 
 export function activate(context: vscode.ExtensionContext): Promise<void> {
 	function registerDisposable<T extends vscode.Disposable>(disposable: T): T {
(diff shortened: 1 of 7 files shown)


src/extension.ts
```diff
@@ -70,8 +71,10 @@ export function activate(context: vscode.ExtensionContext): Promise<void> {
 			telemetryProvider.registerCommandWithTelemetry('vscode-dapr.tasks.scaffoldDaprComponents', createScaffoldDaprComponentsCommand(scaffolder, templateScaffolder));
 			telemetryProvider.registerCommandWithTelemetry('vscode-dapr.tasks.scaffoldDaprTasks', createScaffoldDaprTasksCommand(scaffolder, templateScaffolder, ui));
 			
-			registerDisposable(vscode.tasks.registerTaskProvider('dapr', new DaprCommandTaskProvider(telemetryProvider)));
-			registerDisposable(vscode.tasks.registerTaskProvider('daprd', new DaprdCommandTaskProvider(new NodeEnvironmentProvider(), telemetryProvider)));
+			const settingsProvider = new VsCodeSettingsProvider();
+
+			registerDisposable(vscode.tasks.registerTaskProvider('dapr', new DaprCommandTaskProvider(() => settingsProvider.daprdPath, telemetryProvider)));
```
karolz-ms, line 76:
```suggestion
			registerDisposable(vscode.tasks.registerTaskProvider('dapr', new DaprCommandTaskProvider(() => settingsProvider.daprPath, telemetryProvider)));
```

Not sure... but I think you meant what I am suggesting

karolz-ms (approved):

src/extension.ts
philliphoff: Yes, excellent catch!  I don't know how I managed that.  Fixed.

Issue #165, opened 2021-05-12
Add additional compliance checks during CI


philliphoff:
Our compliance requirements have changed such that we should be doing some additional checks during CI, such as now done by the [vscode-docker](https://github.com/Microsoft/vscode-docker) extension.

PR #168, opened 2021-05-14
Add malware scan step.

philliphoff:
Related to #165.

### DIFF
--- a/.azure-pipelines/compliance/compliance.yml
+++ b/.azure-pipelines/compliance/compliance.yml
@@ -1,4 +1,11 @@
 steps:
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
+  displayName: 'AntiMalware Scanner'
+  inputs:
+    FileDirPath: '$(Build.SourcesDirectory)'
+    EnableServices: true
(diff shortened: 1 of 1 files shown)


.azure-pipelines/compliance/compliance.yml
```diff
@@ -15,6 +15,14 @@ steps:
   continueOnError: true
   condition: in(variables['Build.Reason'], 'Manual', 'Schedule')
 
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
+  displayName: 'AntiMalware Scanner'
+  inputs:
+    FileDirPath: '$(Build.SourcesDirectory)'
+    EnableServices: true
+  continueOnError: true
```
pratiksanglikar, line 23: Do we want to fail the build on error?

philliphoff: That's a good point.  It's borrowed from another VS Code extension pipeline.  The other tasks continue on error in order to get to the post-analysis step, which would then fail if any of the scanners fail or generate a violation. The malware task is supposed to error immediately when finding a problem.  I think I'll move that to the beginning of the tasks and stop right then; no sense continuing a build with detected malware.

pratiksanglikar (approved):

PR #170, opened 2021-05-24
Fixing UnixProcessProvider dapr app match

jasonviviano:
Fixes issue: https://github.com/microsoft/vscode-dapr/issues/169

### DIFF
--- a/src/services/daprApplicationProvider.ts
+++ b/src/services/daprApplicationProvider.ts
@@ -99,7 +99,7 @@ export default class ProcessBasedDaprApplicationProvider extends vscode.Disposab
     }
 
     private async onRefresh(): Promise<void> {
-        const processes = await this.processProvider.listProcesses('daprd');
+        const processes = await this.processProvider.listProcesses('dapr');
         
         this.applications = processes
             .map(process => toApplication(process.cmd, process.pid))


philliphoff:
Unfortunately, this approach does not work with the extension's `daprd` task, which starts `daprd` independently from the `dapr` CLI and, I suspect, when using Dapr with Tye, which does the same.  (Note that the `dapr` CLI itself identifies Dapr applications by existing `daprd` processes and not `dapr` processes.) I think we'll need to parse the full command name from the command line and find those that end with `/daprd`.

philliphoff (changes requested):
Unfortunately, this approach does not work with the extension's `daprd` task, which starts `daprd` independently from the `dapr` CLI and, I suspect, when using Dapr with Tye, which does the same.  (Note that the `dapr` CLI itself identifies Dapr applications by existing `daprd` processes and not `dapr` processes.) I think we'll need to parse the full command name from the command line and find those that end with `/daprd`.

jasonviviano:
Do you think adding just a simple OR condition to the name matching process would then fix this?
Such as instead of just checking for 'daprd' or 'dapr' in the cmd name we can add a conditional so that both dapr cli and also daprd independently initiated apps can be detected?

philliphoff:
The `daprd` process is the de-facto indicator of a running Dapr application, irrespective of who started it (e.g. the `dapr` CLI, VS Code, Tye, etc.), so I think we should focus on that.  Another option is to change how VS Code starts `daprd` to match what the `dapr list` command is looking for (so that it shows up in the list, which it doesn't today) and then use that output rather than look for processes ourselves.  That would push responsibility over to the Dapr side for identifying running applications.  (We'd need to ensure that Tye is also conforming to `list` also.)

jasonviviano:
Sounds good I can look into that, I'm not too familiar with Tye I'm assuming its this project? https://github.com/dotnet/tye

After looking at their implementation, I could be wrong but in my understanding of just reading the code it seems like they look for the daprd path, then try to look into the actual configuration files of Dapr to list out the applications running???
https://github.com/dotnet/tye/blob/652ac4e26677e677d566b61170820f14cd1ebd6a/src/Microsoft.Tye.Extensions/Dapr/DaprExtension.cs#L66

I'm not sure yet what the differences are on how the extension uses daprd to match and what dapr list command uses to match. i'll take a closer look and try to identify what the changes are gonna be and see if it's similar to Tye.

philliphoff:
[Here's](https://github.com/dapr/cli/blob/7dc12c3918a5470f0ec9b922be971096a052fa87/pkg/standalone/list.go#L52) where the Dapr CLI does its matching.

PR #173, opened 2021-05-25
Revise required properties for tasks

### LINKED ISSUE #172: Revise required task properties
We should revise the `dapr` and `daprd` task configuration schemas to reflect the latest set of required properties.  For example, while `command` was required by `dapr` early on, that's no longer required but it's still *marked* as required.  (The workaround for now is to set `command` to `--`.)

philliphoff:
Updates the required properties for the `dapr` and `daprd` tasks based on the latest requirements of the CLI and runtime commands.

Resolves #172.

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,10 +252,7 @@
 						"description": "%vscode-dapr.tasks.dapr.properties.profilePort.description%",
 						"default": -1
 					}
-				},
-				"required": [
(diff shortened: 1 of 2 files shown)


philliphoff:
Oops, forgot to switch the base image. Closing and reopening later.

Issue #81, opened 2020-03-23
Allow killing Dapr applications


philliphoff:
Add an option to kill a running Dapr application (i.e. its `dapr` or `daprd` instance) from the Dapr applications tree view.  If the application does not shut down nominally after debugging, it's not always obvious how/where to stop it so that the user can restart without fear of strange conflicts.  (Especially since the `dapr` CLI does not show arbitrarily started `daprd` instances.)

The applications tree view could offer a "kill" context menu command (at least for instances that it, itself, started) to allow for quick, forced shutdowns and restarts.

jasonviviano:
# Description

## Background
Currently there is no way to kill a Dapr application detected from the UI found in the VS Code extension. Users would have to either kill the debugging process started that has initiated the Dapr application or kill the application from what ever terminal it has been started.

## Request
Add a command that kills a specific Dapr application when that application is selected. This command should be a right click command option that only effects the selected Dapr application.

## Mockup
![Screen Shot 2021-06-21 at 1 46 32 PM](https://user-images.githubusercontent.com/83607984/122805455-1d653a00-d297-11eb-82d7-4ce6c5ffa94c.png)

PR #179, opened 2021-06-09
Implemented issue #81 (allow stopping dapr apps)

sk593:
Implemented the Stop App command to address #81 and added documentation to the readme. There is a bug with the extension not being able to pick up dapr processes (https://github.com/microsoft/vscode-dapr/issues/169) so testing this will require a temporary fix similar to whats proposed here (https://github.com/microsoft/vscode-dapr/pull/170) 

@philliphoff @pkedy @artursouza


src/services/daprClient.ts
```diff
@@ -60,4 +63,14 @@ export default class HttpDaprClient implements DaprClient {
 
         await this.httpClient.post(url, payload, { json: true }, token);
     }
-}
\ No newline at end of file
+
+    async stopApp(application: DaprApplication): Promise<void> {
+        const cmdResponse = await Process.exec(`dapr stop --app-id ${application.appId}`);
```
philliphoff, line 68: A couple of comments:

1. Use the `daprPathProvider` pattern to locate `dapr` rather than assume it's in the `PATH`.
1. Consider using the `CommandLineBuilder` for building the path to ensure quoting, if necessary.
1. Does `dapr stop` work when `daprd` is started directly, or only when `dapr` was used to start the sidecar and the app?

src/services/daprClient.ts
```diff
@@ -60,4 +63,14 @@ export default class HttpDaprClient implements DaprClient {
...
+        
+        if(cmdResponse.code == 0) {
+            void vscode.window.showInformationMessage(cmdResponse.stdout);
```
philliphoff, line 71: Should we display a message when successful, or only when it fails?  (VS Code tends to bombard users with notifications, so we try to limit them only to when absolutely necessary.)

In any case, we should use the `UserInput` interface to show messages (which is an abstraction over the VS Code API).

src/services/daprClient.ts
artursouza: `dapr stop` relies on the output from `dapr list` - which only works if daprd is started from dapr CLI - it is because we rely on the params from dapr CLI process to track information about the sidecar and corresponding app.

sk593: Are there specific parameters that are required by the CLI for a daprd process to show up in dapr list (i.e. any daprd call needs to have an app id, dapr http port, etc for it to be picked up by dapr list)?

philliphoff: It looks like the `dapr list` implementation expects to see the following `daprd` arguments:

 - `--dapr-http-port`
 - `--dapr-grpc-port`
 - `--dapr-http-max-request-size`

The last one should arguably *not* be required; we should file a bug for that, but it should be easy enough to add to the extension's invocations of `daprd`.

src/commands/applications/invokeCommon.ts
```diff
@@ -143,5 +143,10 @@ export async function invoke(context: IActionContext, daprApplicationProvider: D
 }
 
 export async function stop(context: IActionContext, daprApplicationProvider: DaprApplicationProvider, daprClient: DaprClient, ui: UserInput, application: DaprApplication | undefined): Promise<void> {
-    await daprClient.stopApp(await getApplication(context.telemetry, daprApplicationProvider, ui, application))
+    try {
+        daprClient.stopApp(await getApplication(context.telemetry, daprApplicationProvider, ui, application))
+    } catch(error) {
+        await ui.showWarningMessage(localize('commands.invokeCommon.stopAppError', 'Failed to stop the selected application'),
```
philliphoff, line 149: As VS Code notifications tend to pile up, it'd probably be useful to include the actual name of the application in the error message (rather than just "the selected application").

package.nls.json
```diff
@@ -3,6 +3,7 @@
     "vscode-dapr.applications.invoke-post.title": "Invoke (POST) Application Method",
     "vscode-dapr.applications.publish-message.title": "Publish Message to Application",
     "vscode-dapr.applications.publish-all-message.title": "Publish Message to All Applications",
+    "vscode-dapr.applications.stop-app.title": "Stop Selected Application",
```
philliphoff, line 6: I'd suggest "Stop Application" (as it's somewhat assumed that it's referring to the currently selected one).

src/commands/applications/invokeCommon.ts
```diff
@@ -140,4 +140,13 @@ export async function invoke(context: IActionContext, daprApplicationProvider: D
 
             outputChannel.show();
         });
+}
+
+export async function stop(context: IActionContext, daprApplicationProvider: DaprApplicationProvider, daprClient: DaprClient, ui: UserInput, application: DaprApplication | undefined): Promise<void> {
```
philliphoff, line 145: It's probably not necessary to put `stop()` in `invokeCommon.ts`, as it's not sharing any of the same wizard UX that the other invocation methods were sharing (and why they were grouped together).

src/services/daprClient.ts
```diff
@@ -60,4 +62,8 @@ export default class HttpDaprClient implements DaprClient {
 
         await this.httpClient.post(url, payload, { json: true }, token);
     }
-}
\ No newline at end of file
+
+    stopApp(application: DaprApplication): void {
+        process.kill(application.pid, 'SIGKILL')
```
philliphoff, line 67: I would think it better to use `SIGTERM` (which is the default), which would give Dapr the opportunity to shutdown the application and sidecar more gracefully than with `SIGKILL`.  Note that Windows doesn't support the standard set of signals nor is consistent in its behavior; in the [`vscode-tye` extension we use `taskkill` instead](https://github.com/microsoft/vscode-tye/blob/291fc36e741fcd414b692ab224e5ecc15103a172/src/util/process.ts#L151) (which will kill the entire process tree).

src/commands/applications/invokeCommon.ts
sk593: I had it in `invokeCommon.ts` so that the `getApplications()` method could be reused to find the application to be stopped. Would it be better to move the logic to be in `daprClient.ts`?

philliphoff: If it's using the same logic as the other functions, then it's fine where it is.

src/services/daprClient.ts
sk593: I updated the function to include the support for Windows. Using `SIGTERM` only seems to kill instances of `daprd`, but not `dapr` processes run through the CLI (i.e. through a `dapr run`). It seems like the signal might be getting intercepted somewhere but `SIGKILL` is able to shutdown both the app and sidecar (and an independent sidecar).

philliphoff: In the case of a `dapr` started  `daprd` sidecar and application, which process is the one getting killed?  We may need to create a table that shows the various scenarios (e.g. `dapr` with sidecar and app, `dapr` with just sidecar, `daprd` sidecar only) so that we understand what we expect the stop command to do in each case.  I'm still hesitant to use `SIGKILL` without first trying `SIGTERM`, but it may only work if we're targeting the correct process.  It may require listing both `dapr` and `daprd` processes separately and then correlating them to understand which process to target.

sk593: I went through the different scenarios and this seems to be the results:

1. dapr sidecar only
    1. Killing the process of the daprd command kills the process
2. dapr with app and sidecar
    1. Killing just the dapr process using `SIGTERM` kills both daprd sidecar and the aren’t dapr process
    2. Killing just the daprd sidecar process using `SIGTERM` kills both the daprd sidecar and the parent dapr process 
3. dapr with just sidecar
    1. Killing dapr process with `SIGTERM` kills sidecar and run process 
    2. Killing daprd process with `SIGKILL` kills sidecar and run process

Since the processes that we find in the processProvider are `daprd` processes, I've been killing `daprd` processes by pid and this can kill both the sidecar and the app (but this only covered all cases when using `SIGKILL` since `SIGTERM` only covered certain scenarios). However, based on the different scenarios `SIGTERM` can be used if we keep track of the `ppid` for any `daprd` process that's found. That way, in scenarios 2/3, we can kill the parent `dapr` process in order to kill the process tree.

src/services/processProvider.ts
```diff
@@ -50,8 +65,9 @@ export class WindowsProcessProvider implements ProcessProvider {
             const cmd = getWmicValue(lines[(i * 5) + 2]);
             const name = getWmicValue(lines[(i * 5) + 3]);
             const pid = parseInt(getWmicValue(lines[(i * 5) + 4]), 10);
+            const ppid = undefined;
```
philliphoff, line 68: If we don't get the parent process on Windows, does that mean it will have different behavior in some scenarios? It *looks* like we can get the parent process via WMIC by just adding `parentprocessid` to the list of returned columns.

src/services/processProvider.ts
```diff
@@ -18,11 +19,25 @@ export interface ProcessProvider {
...
     async listProcesses(name: string): Promise<ProcessInfo[]> {
         const processes = await psList();
+        const temp = processes
+            .filter(process => process.name === name || this.hasDaprdPath(process))
+            .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid , ppid: this.getDaprPpid(process)}));
+        return temp;
+    }
+    
+    hasDaprdPath(process: psList.ProcessDescriptor): boolean {
+        const cmdPath = process.cmd?.split(" ")[0];
```
philliphoff, line 29: This will fail if the `daprd` binary exists along a path that contains whitespace.  This is likely (in particular, on Windows) if the user places the binary in a user-specific path, as they are often proper names (e.g. `"C:\Users\Phil Hoff\.dapr\bin\daprd.exe`").

src/services/processProvider.ts
```diff
@@ -18,11 +19,25 @@ export interface ProcessProvider {
...
+        return cmdPath?.indexOf("/") != -1 //check that contains daprd filepath and not daprd cmd
```
philliphoff, line 30: Nit: `!==`

src/services/daprClient.ts
```diff
@@ -71,7 +71,12 @@ export default class HttpDaprClient implements DaprClient {
             //       As a partial mitigation, use `taskkill` to kill the entire process tree.
             void Process.exec(`taskkill /pid ${application.pid} /t /f`);
         } else {
-            process.kill(application.pid, 'SIGKILL')
+            if(application.ppid != undefined) {
```
philliphoff, line 75: Suggestion:

```typescript
process.kill(application.ppid !== undefined ? application.ppid : application.pid, 'SIGTERM');
```

src/services/daprClient.ts
```diff
@@ -71,7 +71,12 @@ export default class HttpDaprClient implements DaprClient {
             //       As a partial mitigation, use `taskkill` to kill the entire process tree.
             void Process.exec(`taskkill /pid ${application.pid} /t /f`);
```
philliphoff, line 73: If we track the parent process for Windows, too, then we'd want the same switch between process/parent here, too.

src/services/processProvider.ts
sk593: I had tested the kill process on Windows and killing the `daprd` processs kills both the `daprd` process and the parent `dapr` process so there wasn't technically a need for the parent process id. But I can add it so that the logic is the same between different OS's.

src/services/daprClient.ts
```diff
@@ -62,6 +67,22 @@ export default class HttpDaprClient implements DaprClient {
         await this.httpClient.post(url, payload, { json: true }, token);
     }
 
+    async stopApp(application: DaprApplication | undefined, ui: UserInput): Promise<void> {
```
philliphoff, line 70: Should we allow a `undefined` `application` to be passed in?  What would the expected behavior be?

src/services/daprClient.ts
```diff
@@ -62,6 +67,22 @@ export default class HttpDaprClient implements DaprClient {
...
+        try {
+            if (os.platform() === 'win32') {
+                // NOTE: Windows does not support SIGTERM/SIGINT/SIGBREAK, so there can be no graceful process shutdown.
+                //       As a partial mitigation, use `taskkill` to kill the entire process tree.
+                processId !== undefined ? void Process.exec(`taskkill /pid ${processId} /t /f`) : null;
+            } else {
+                processId !== undefined ? process.kill(processId, 'SIGTERM') : null;
+            } 
+        } catch(error) {
+            await ui.showWarningMessage(localize('commands.invokeCommon.stopAppError', 'Failed to stop application \'{0}\'', application?.appId),
```
philliphoff, line 81: There is some default error handling for commands that fail, i.e. throw an exception--it's part of the telemetry wrapping.  That may be sufficient for our needs; if not, then the outer command probably ought to handle the modal dialog rather than the `DaprClient`.

src/services/processProvider.ts
```diff
@@ -18,11 +21,25 @@ export interface ProcessProvider {
...
+        const daprdPath = which.sync('daprd', {nothrow: true});
+        const executable = daprdPath !== null ? daprdPath : which.sync('daprd.exe', {nothrow: true});
+        return executable !== null ? (process.cmd?.indexOf(executable) !== -1) : false;
+    }
 
-        return processes
-            .filter(process => process.name === name)
-            .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid }));
+    getDaprPpid(process: psList.ProcessDescriptor): number | undefined {
+        if(this.hasDaprdPath(process)) {
```
philliphoff, line 37: In what case would this *not* be true?  (I.e. it may have already been called by the prior `filter()`.)

src/services/daprClient.ts
sk593: I believe that it will almost never be the case where the application is undefined since you have to choose a node to stop. However, the input types for the telemetry provider ask for `node | undefined` (for the DaprApplicationNode) so I carried it through to the `DaprClient`. I can update the code to stop the function within the `stopApp` class if the application node ends up being undefined, if that's preferred

src/services/processProvider.ts
sk593: My bad, this was from the previous proc bug fix commit where it was needed to ensure that processes of the sidecar running independently wouldn't be registered as having a parent `dapr` process. The conditional most likely will not be needed with any new changes to the proc bug fix.

src/services/daprClient.ts
sk593: I don't think the telemetry provider gives any UI feedback when commands fail, just an output in the console. I think a UI message would be preferred as a form of feedback and can move it to the outer function, but if its not necessary then I can remove the error handling

src/services/daprInstallationManager.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
philliphoff, line 1: Nit: we prefer to disable linting rules for individual lines rather than whole files.  (And there are often tricks to make the TS compiler happy, than can avoid having to disable certain kinds of rules at all.)

philliphoff (approved):
A couple of comments, and then I think it can go in after your related `daprd` path changes are merged.

Issue #159, opened 2021-05-10
Add link to Dapr dashboard from Explorer


BigMorty:

jasonviviano:
## Description
Add the ability to initiate the Dapr Dashboard from the extension. Solutions looked into are both a dashboard icon in the Application tree view and a command palette function that both use up a unique free port and start up the Dapr Dashboard in a user's browser.

sk593:
Closed in PR #190

PR #180, opened 2021-06-14
Implemented issue #159 Dapr Dashboard

jasonviviano:
## Issue

https://github.com/microsoft/vscode-dapr/issues/159

## Description

Added a tree view "Open Dapr Dashboard" command and a command palette "Open Dapr Dashboard" Command.

Command opens up a new terminal called "Dapr Dashboard Terminal", runs command "dapr dashboard -p ${port}" in where port is a unique open port. Command displays information message stating terminal already exists if terminal is still open and command and is run again.

## Tests

Tested out manually, both new UI icon displayed on "Applications" tree view and command palette.

## UI
Screen Shot 2021-06-22 at 2 02 01 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
 				"command": "vscode-dapr.tasks.openDaprDashboard",
 				"title": "%vscode-dapr.tasks.openDaprDashboard.title%",
 				"category": "Dapr",
-				"icon": "$(default-view-icon)"
+				"icon": "$(dashboard)"
 			}
 		],
 		"configuration": {
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
 		"onCommand:vscode-dapr.tasks.scaffoldDaprComponents",
 		"onCommand:vscode-dapr.tasks.scaffoldDaprTasks",
 		"onCommand:workbench.action.tasks.runTask",
+		"onCommand:vscode-dapr.tasks.openDaprDashboard",
 		"onView:vscode-dapr.views.applications"
 	],
 	"repository": {
@@ -98,6 +99,12 @@
 				"command": "vscode-dapr.tasks.scaffoldDaprTasks",
 				"title": "%vscode-dapr.tasks.scaffoldDaprTasks.title%",
 				"category": "Dapr"
+			},
+			{
+				"command": "vscode-dapr.tasks.openDaprDashboard",
+				"title": "%vscode-dapr.tasks.openDaprDashboard.title%",
+				"category": "Dapr",
+				"icon": "$(default-view-icon)"
 			}
 		],
 		"configuration": {
@@ -131,6 +138,11 @@
 			"view/title": [
 				{
 					"command": "vscode-dapr.applications.publish-all-message",
+					"group": "navigation",
+					"when": "view == vscode-dapr.views.applications"
+				},
+				{
+					"command": "vscode-dapr.tasks.openDaprDashboard",
 					"group": "navigation",
 					"when": "view == vscode-dapr.views.applications"
 				}
@@ -470,6 +482,7 @@
 		"axios": "^0.21.1",
 		"fs-extra": "^9.1.0",
 		"handlebars": "^4.7.7",
+		"portfinder": "^1.0.28",
 		"ps-list": "^7.2.0",
 		"vscode-azureextensionui": "^0.40.0",
 		"vscode-nls": "^5.0.0"
--- a/src/commands/openDaprDashboard.ts
+++ b/src/commands/openDaprDashboard.ts
@@ -1,35 +1,47 @@
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/no-floating-promises */
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
 import * as vscode from 'vscode';
 import portfinder = require('portfinder');
+import { spawn } from 'child_process';
+import createPlatformProcessProvider from '../services/processProvider';
 
 
+function getPortFromCommand(command: string): string {
+    if (command.includes('--port')){
+        command = command.split('--port')[1].toString().trim().split(" ")[0]
+    }
+    if (command.includes('-p') ){
+        command = command.split('-p')[1].toString().trim().split(" ")[0]
+    }
+    return command;
+}
+class ProcessBasedDaprDashboardProvider {
+    private processPlatform = createPlatformProcessProvider();
+    port:string | undefined;
 
-function ensureTerminalDoesNotExist(): boolean {
-    const terminals = <vscode.Terminal[]>(<any>vscode.window).terminals;
-    for (const terminal of terminals) {
-        if (terminal.name === 'Dapr Dashboard Terminal') {
-            vscode.window.showInformationMessage('Dapr Dashboard Terminal Already Exists');
-            return false;
+    async checkForDaprDashboard(): Promise<void> {
+        const dashboardProcesses = await this.processPlatform.listProcesses('dashboard');
+        if (dashboardProcesses.length > 0) {
+            this.port = getPortFromCommand(dashboardProcesses[0].cmd)
         }
     }
-	return true;
 }
-
 export async function openDaprDashboard(): Promise<void> {
-    if (ensureTerminalDoesNotExist()){
-        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
-        await portfinder.getPortPromise()
+    const daprDashboardProvider = new ProcessBasedDaprDashboardProvider();
+
+    await daprDashboardProvider.checkForDaprDashboard();
+
+    if (daprDashboardProvider.port == undefined){
+        portfinder.getPortPromise()
         .then((port) => {
-            const portUsed: number = port;
-            terminal.sendText(`dapr dashboard -p ${portUsed}`);
-            vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(`http://localhost:${portUsed}`));
-        })
+            spawn('dapr', ['dashboard', '-p', `${port}`]);
+            daprDashboardProvider.port = port.toString();
+        }) 
         .catch((err) => {
-            return Promise.reject(err);
-        });
-    }
+            console.log(err);
+        });    } 
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    await vscode.env.openExternal(vscode.Uri.parse(`http://localhost:${daprDashboardProvider.port}`));
 }
 
 const createOpenDaprDashboardCommand = () => (): Promise<void> => openDaprDashboard();
--- a/src/commands/openDaprDashboard.ts
+++ b/src/commands/openDaprDashboard.ts
@@ -1,14 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 import * as vscode from 'vscode';
-import DaprDashboardProvider from '../services/daprDashboardProvider';
+import ClassBasedDaprDashboardProvider from '../services/daprDashboardProvider';
 
 
-export async function openDaprDashboard( daprDashboardProvider: DaprDashboardProvider): Promise<void> {
+export async function openDaprDashboard( classBasedDaprDashboardProvider: ClassBasedDaprDashboardProvider): Promise<void> {
 
     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-    await vscode.env.openExternal(vscode.Uri.parse(`http://localhost:${daprDashboardProvider.getPortUsed()}`));
+    await vscode.env.openExternal(await classBasedDaprDashboardProvider.startDashboard());
 }
 
-const createOpenDaprDashboardCommand = ( daprDashboardProvider: DaprDashboardProvider) => (): Promise<void> => openDaprDashboard( daprDashboardProvider);
+const createOpenDaprDashboardCommand = ( classBasedDaprDashboardProvider: ClassBasedDaprDashboardProvider) => (): Promise<void> => openDaprDashboard( classBasedDaprDashboardProvider);
 export default createOpenDaprDashboardCommand;
--- a/src/commands/openDaprDashboard.ts
+++ b/src/commands/openDaprDashboard.ts
@@ -1,0 +1,36 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import * as vscode from 'vscode';
+import portfinder = require('portfinder');
+
+
+
+function ensureTerminalDoesNotExist(): boolean {
+    const terminals = <vscode.Terminal[]>(<any>vscode.window).terminals;
+    for (const terminal of terminals) {
+        if (terminal.name === 'Dapr Dashboard Terminal') {
+            vscode.window.showInformationMessage('Dapr Dashboard Terminal Already Exists');
+            return false;
+        }
+    }
+	return true;
+}
+
+export async function openDaprDashboard(): Promise<void> {
+    if (ensureTerminalDoesNotExist()){
+        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
+        await portfinder.getPortPromise()
+        .then((port) => {
+            const portUsed: number = port;
+            terminal.sendText(`dapr dashboard -p ${portUsed}`);
+            vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(`http://localhost:${portUsed}`));
+        })
+        .catch((err) => {
+            return Promise.reject(err);
+        });
+    }
+}
+
+const createOpenDaprDashboardCommand = () => (): Promise<void> => openDaprDashboard();
+export default createOpenDaprDashboardCommand;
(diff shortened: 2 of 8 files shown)


philliphoff:
@jasonviviano Can you add a screenshot showing the new UX (which often helps when reviewing the source).

package.json
```diff
@@ -98,6 +99,12 @@
 				"command": "vscode-dapr.tasks.scaffoldDaprTasks",
 				"title": "%vscode-dapr.tasks.scaffoldDaprTasks.title%",
 				"category": "Dapr"
+			},
+			{
+				"command": "vscode-dapr.tasks.openDaprDashboard",
+				"title": "%vscode-dapr.tasks.openDaprDashboard.title%",
+				"category": "Dapr",
+				"icon": "$(default-view-icon)"
```
philliphoff, line 107: For `vscode-tye`, we're using the `dashboard` icon.  We probably want to be consistent across extensions, whichever is the better icon.

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
philliphoff, line 1: Nits:

 - Need the standard copyright header.
 - We shouldn't globally disable lint rules; only disable them for single lines at a time (and try to avoid the need to disable them at all, if possible).

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export async function openDaprDashboard(): Promise<void> {
+    if (ensureTerminalDoesNotExist()){
+        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
+        await portfinder.getPortPromise()
+        .then((port) => {
+            const portUsed: number = port;
+            terminal.sendText(`dapr dashboard -p ${portUsed}`);
+            vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(`http://localhost:${portUsed}`));
```
philliphoff, line 27: Prefer to use the `UserInput.openExternal()` service method to open the browser (which is an abstraction on the VS Code APIs to do so).

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
...
+            vscode.window.showInformationMessage('Dapr Dashboard Terminal Already Exists');
+            return false;
+        }
+    }
+	return true;
+}
+
+export async function openDaprDashboard(): Promise<void> {
+    if (ensureTerminalDoesNotExist()){
+        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
```
philliphoff, line 22: Do we need to open a terminal window for this?  Does the user need to see its output?  Can we just use `Process.spawn()`?

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
...
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import * as vscode from 'vscode';
+import portfinder = require('portfinder');
+
+
+
+function ensureTerminalDoesNotExist(): boolean {
+    const terminals = <vscode.Terminal[]>(<any>vscode.window).terminals;
+    for (const terminal of terminals) {
+        if (terminal.name === 'Dapr Dashboard Terminal') {
```
philliphoff, line 12: I believe the user can rename terminal windows now, so I don't know if we should count on that to find a running dashboard.  In any case, I don't think we should display an error if it already exists.  (The user doesn't care whether there's an existing dashboard or not, they just want to browse to one.)

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
...
+	return true;
+}
+
+export async function openDaprDashboard(): Promise<void> {
+    if (ensureTerminalDoesNotExist()){
+        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
+        await portfinder.getPortPromise()
+        .then((port) => {
+            const portUsed: number = port;
+            terminal.sendText(`dapr dashboard -p ${portUsed}`);
```
philliphoff, line 26: It'd be great if we could shift responsibility for finding an open port to `dapr dashboard`.  I filed a related issue dapr/cli#743.  Until then, however, we likely don't have a choice.

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+	return true;
+}
+
+export async function openDaprDashboard(): Promise<void> {
+    if (ensureTerminalDoesNotExist()){
+        const terminal = vscode.window.createTerminal(`Dapr Dashboard Terminal`);
+        await portfinder.getPortPromise()
+        .then((port) => {
```
philliphoff, line 24: Nit we should try to avoid mixing `async/await` style and `then()` style Promises.

philliphoff (changes requested):
I think we'll want to hide the dashboard process from the user rather than relying on terminal windows (which the user has the ability to manipulate).

src/commands/openDaprDashboard.ts
jasonviviano: Opening up the terminal allows me to guarantee that we won't be taking up any ports away if the user exits VS Code. I'm not too familiar if process.spawn() operates differently from just creating a child process and letting it run, since when i tried that it's behavior was a bit less consistent in terminating.

jasonviviano: Do you think that would work better?

src/commands/openDaprDashboard.ts
jasonviviano: I was using it just as a check to not open multiple dashboards constantly. I'm not sure if we want to limit the amount of times the user can call this function if its already running?

src/commands/openDaprDashboard.ts
philliphoff: VS Code has a notion of registering "disposable" components during extension activation.  It will then call the `dispose()` method of each registered component when the extension is deactivated (such as when VS Code is shut down).  I'd suggest using that as the trigger for killing the process spawned for the dashboard.

src/commands/openDaprDashboard.ts
philliphoff: Agree that we don't want to open more than one dashboard at any given time, but we should be able to do that without relying on a terminal (or its name) to do that (because terminals are subject to manipulation by users).

jasonviviano:
Added a process check to find already running instances of the dashboard and if not command opens its on

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,48 @@
...
+        const dashboardProcesses = await this.processPlatform.listProcesses('dashboard');
+        if (dashboardProcesses.length > 0) {
+            this.port = getPortFromCommand(dashboardProcesses[0].cmd)
+        }
+    }
+}
+export async function openDaprDashboard(): Promise<void> {
+    const daprDashboardProvider = new ProcessBasedDaprDashboardProvider();
+
+    await daprDashboardProvider.checkForDaprDashboard();
```
philliphoff, line 32: I'm not sure why we're concerned with re-use of an existing dashboard.  We always have to handle the case where it's not yet running, in which case we have to manage starting it (and stopping it when VS Code is closed).  Isn't *just* doing the latter actually simpler?

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (daprDashboardProvider.port == undefined){
+        portfinder.getPortPromise()
+        .then((port) => {
+            spawn('dapr', ['dashboard', '-p', `${port}`]);
```
philliphoff, line 37: The user can set an explicit path to the `dapr` executable, which should be respected when starting the dashboard (e.g. via a passed in `settingsProvider`).

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,48 @@
...
+        command = command.split('-p')[1].toString().trim().split(" ")[0]
+    }
+    return command;
+}
+class ProcessBasedDaprDashboardProvider {
+    private processPlatform = createPlatformProcessProvider();
+    port:string | undefined;
+
+    async checkForDaprDashboard(): Promise<void> {
+        const dashboardProcesses = await this.processPlatform.listProcesses('dashboard');
```
philliphoff, line 23: Is the dashboard a separate executable from the `dapr` CLI executable?  If so, we'll likely have similar concerns with how other processes have started it (e.g. via its name or its full path), which has similar concerns to what @sk593 is working on with regard to listing `daprd` processes.

jasonviviano: I believe so it has its own repo https://github.com/dapr/dashboard

src/commands/openDaprDashboard.ts
jasonviviano: So if its running outside of the VS Code instance we are just opening the tab in the browser, when we start it from VS Code it will close once the VS Code instance is closed based on my testing.

src/commands/openDaprDashboard.ts
jasonviviano: I'm not sure I understand what you mean here. You want the user to pass their dapr path?

jasonviviano: Ah so make use of the settingsProvider to grab the dapr path for the spawn?

philliphoff: Yes, exactly.  (The `settingsProvider` would be passed to the dashboard command/manager.)

src/commands/openDaprDashboard.ts
philliphoff: Yes, but why go to the effort of re-using an existing dashboard if we already have logic for starting/stopping it (which we have to have, anyway)?  Why not just *always* start the dashboard.  There seems to be little reason *not* to.

Also, you'll want to verify that the dashboard is closed on each platform; Windows may act differently with respect to child processes than Mac/Linux.  It's usually better to actively manage the shutdown of helper processes rather than relying on the system to clean up on our behalf.

jasonviviano: There's not real way to stop the dashboard without ctrl-c to stop the process. There's also no difference between any dashboard instance when starting one. If one is already running on the machine I think its better to reuse it since that means the user is already aware of the dashboard and the port they are using so it gives them greater control.


I'll verify on Windows after the standup! Make sure it reacts the same with each behavior.

philliphoff: I don't think it's a given that the user is aware of a pre-existing dashboard.  Suppose the users goes to the dashboard from VS Code and it's not already running (and so is started within the context of the VS Code session).  The user then opens another VS Code window and also goes to the dashboard.  The dashboard is currently running, and so the browser simply opens.  The user then closes the first VS Code window; as it started the dashboard, it will result in its shutdown.  But the user has no real way of knowing that it's been shutdown; she may still expect the dashboard to work (because she browsed to it from the second VS Code window, which is still open).

I think the most better experience would be for each VS Code window to use its own dashboard instance, as that provides the most consistent conceptual model of the lifetime of the dashboard.  That is, that the dashboard will be functional, at least, for as long as the VS Code window that the user opens it from remains open.

jasonviviano: Got it! Thanks for the feedback have updated to only use a VS Code instance of dashboard.

jasonviviano:
Created a class that once instantiated is the dashboard data hub that the command interacts with @philliphoff let me know if this is more to your liking!

src/extension.ts
```diff
@@ -70,9 +72,11 @@ export function activate(context: vscode.ExtensionContext): Promise<void> {
 			telemetryProvider.registerContextCommandWithTelemetry('vscode-dapr.help.reviewIssues', createReviewIssuesCommand(ui));
 			telemetryProvider.registerCommandWithTelemetry('vscode-dapr.tasks.scaffoldDaprComponents', createScaffoldDaprComponentsCommand(scaffolder, templateScaffolder));
 			telemetryProvider.registerCommandWithTelemetry('vscode-dapr.tasks.scaffoldDaprTasks', createScaffoldDaprTasksCommand(scaffolder, templateScaffolder, ui));
-			
 			const settingsProvider = new VsCodeSettingsProvider();
 
+			const daprDashboardProvider = new DaprDashboardProvider(settingsProvider.daprPath);
```
philliphoff, line 77: The `dapr` path may change over time, for example, if the user modifies the setting after the extension has been activated.  Therefore, we can't rely on the setting at the time of activation.  Instead, consider having `DaprDashboardProvider` accept a function that, when called, returns the current setting (e.g. `daprPathProvider: () => string`).

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT license.
+import portfinder = require('portfinder');
+import { spawn } from 'child_process';
+
+export default class DaprDashboardProvider {
+    private port:number | undefined;
+    private daprPath: string
+
+    constructor(daprPath: string) {
+        portfinder.getPortPromise()
```
philliphoff, line 11: This would cause the dashboard to be started immediately upon construction (i.e. upon extension activation) which is probably not what we want; I think we'd want to defer creation of the dashboard until the user actually requests it for the first time.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.daprPath = daprPath;
+    }
+
+    spawnDashboardInstance(): void {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        spawn(this.daprPath, ['dashboard', '-p', `${this.port}`]);
+    }
+
+    getPortUsed(): number | undefined {
+        return this.port;
```
philliphoff, line 32: There is a race condition here in that the port may not have been determined (because it's happening asynchronously) when this function is called.  Instead, it would probably be better to have an interface more like this:

```typescript
interface DaprDashboardProvider {
    /*
      Starts the Dapr dashboard (if not already started) and returns the URI at which it's running.
    */
    startDashboard(): Promise<vscode.Uri>;
}
```

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+            console.log(err);
+            this.port = 8080;
+            this.spawnDashboardInstance()
+        });
+        this.daprPath = daprPath;
+    }
+
+    spawnDashboardInstance(): void {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        spawn(this.daprPath, ['dashboard', '-p', `${this.port}`]);
```
philliphoff, line 28: I suggest using the `Process` wrapper around `child_process` in the `util` folder.  It would also allow the provider to hold a cancellation token with which it can shutdown the dashboard when the provider itself is disposed.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+    constructor(daprPath: string) {
+        portfinder.getPortPromise()
+        .then((port) => {
+            this.port = port;
+        })
+        .then(() => {
+            this.spawnDashboardInstance()
+        })
+        .catch((err) => {
+            console.log(err);
```
philliphoff, line 19: If there's a failure obtaining a dynamic port or starting the dashboard on that port, is there a particular reason to try with port 8080?  (And what happens if that, also, fails?)

jasonviviano: Port 8080 is the default port the application starts in so i figured it would be a good backup, however I also don't believe it should fail the code in the library seems to be stable that it shouldn't and if it does it doesn't break the extension outside of the feature.

jasonviviano: I'll set it to a obscure number port to avoid conflicts since 8080 is common.

src/services/daprDashboardProvider.ts
jasonviviano: I tried something out but let me know if I'm missing the point with the interface, I'm not too familiar with using them.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    async findOpenPort(): Promise<void> {
+        await portfinder.getPortPromise()
+        .then((port) => {
+            this.port = port;
+        })
+        .catch((err) => {
+            console.log(err);
+            this.port = 50505;
```
philliphoff, line 41: If we can't find an open port, I don't think trying a static port would make sense (as presumably that one is already in use).  In that unlikely case, we should just throw an error.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+import {Process} from '../util/process'
+export interface DaprDashboardProvider {
+    startDashboard(): Promise<vscode.Uri>;
+}
+
+export default class ClassBasedDaprDashboardProvider implements DaprDashboardProvider {
+    private port:number | undefined;
+    private daprPath: string
+    private process = new Process();
+    constructor(daprPath: string) {
```
philliphoff, line 14: The Dapr path may change over time, so this class should accept a function that returns the dapr path (on each call), rather than a static value.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+            return;
+        }
+        await this.findOpenPort();
+        await this.processSpawn();
+    }
+
+    async findOpenPort(): Promise<void> {
+        await portfinder.getPortPromise()
+        .then((port) => {
+            this.port = port;
```
philliphoff, line 37: If we're not able to start the actual dashboard process (after getting a valid port), we won't be able to startup the dashboard later (until VS Code is reopened) because we'll already have a port (which was only valid at the first try, but may not be valid on the next try anyway).

sk593:
@jasonviviano Is it okay if I pick up this PR? We want to have it in the release

jasonviviano:
> @jasonviviano Is it okay if I pick up this PR? We want to have it in the release

Yeah no worries @sk593

sk593:
Closing this. See #190 for new PR and updated changes

Issue #169, opened 2021-05-24
UnixProcessProvider Failed to Find Dapr Process CMD on Mac


jasonviviano:
**Description**
The current implementation for finding Dapr processes uses the library ps-list, the usage can be found here:
https://github.com/microsoft/vscode-dapr/blob/665dabfc384404d95292a209cd692c726277def2/src/services/processProvider.ts#L23

There's a method for Windows and Mac, however on Mac there's been a bug were the current Dapr applications have trouble being found. It seems that the current processes cmd name matching relied on 'daprd' but simply changing the name to 'dapr' as the matching string allowed for dapr cli initiated apps to be found.

** Tests **
Tested locally with running the following Dapr apps, 

dapr run --app-id pppptest --app-port 3560 --dapr-http-port 3850 node app.js
dapr run --app-id nodeapptesttt --app-port 3550 --dapr-http-port 3800 node app.js

PR #185, opened 2021-06-18
Bug fix for daprd proc bug #169

sk593:
Bug fix for #169. Updated application finder to locate all daprd processes (run by either the dapr cli or through daprd directly). 

@philliphoff @philkedy


src/services/daprApplicationProvider.ts
```diff
@@ -99,12 +100,23 @@ export default class ProcessBasedDaprApplicationProvider extends vscode.Disposab
     }
 
     private async onRefresh(): Promise<void> {
-        const processes = await this.processProvider.listProcesses('daprd');
```
philliphoff, line 102: Any fix probably ought to be incorporated within the `ProcessProvider`, if anything, because the `ps-list` library doesn't generate a `cmd` property on Windows so this fix wouldn't work on that platform. To a certain extent, `ProcessProvider` exists to hide the switching between `ps-list` on Linux/Mac and `wmic` on Windows.

src/services/daprApplicationProvider.ts
```diff
@@ -99,12 +100,23 @@ export default class ProcessBasedDaprApplicationProvider extends vscode.Disposab
...
-        
-        this.applications = processes
-            .map(process => toApplication(process.cmd, process.pid))
+        const obj = await psList();
+        this.applications = obj
+            .map((process: psList.ProcessDescriptor) => process as ProcessInfo)
+            .filter((process: ProcessInfo) => (process.cmd != undefined && process.cmd.indexOf("daprd") != -1))
```
philliphoff, line 106: This would also include command lines that just happened to have the string `daprd` somewhere within it, not just invocations of `daprd`.  The command line we're looking for is something like `/User/dapr/bin/daprd --app-id my-app ...`.  We should only be looking for `daprd` filename at the end of the command name and ignore any following arguments.

sk593: When listing out the processes that invoke daprd (i.e. using something like ps -ef | grep dapr/daprd) the daprd process run through the VS Code extension doesn't have the `daprd` filename so the process is only listed as something like `daprd --app-id my-app...` without the `daprd` at the end of the command name. So searching for the filename won't return an app process started by the extension. Would it be better then to use the process provider, but search for both instances of `dapr` and `daprd` instead of just `daprd` as we are right now?

src/services/daprApplicationProvider.ts
sk593: It seemed to work when I tested it on Windows, but it's possible I was using a Linux subsystem and that's why it worked. But I'll move the changes to the ProcessProvider to keep the implementation wrapped in that class

src/services/daprApplicationProvider.ts
philliphoff: I think we'll want to look for both `daprd --app-id...` and `/User/.dapr/bin/dapr --app-id...` forms, as the latter would also used if the user configures an explicit `daprd` path.  As long as the sidecar can be started independently from `dapr`, looking for those processes really represents the source of truth.  (At least until we can relegate responsibility to `dapr list`, now that dapr/cli#729 has been committed, and are able to rip out the process provider.)

src/services/daprApplicationProvider.ts
philliphoff: Yeah, using VS Code with WSL would end up using the Linux codepaths.  I use Parallels on my Mac in order to setup both a Windows and Linux VM, so that I can easily test across the different platforms.

src/services/daprApplicationProvider.ts
sk593: Does the linked issue change any of the CLI list implementation (i.e. any of the logic in [this file](https://github.com/dapr/cli/blob/master/pkg/standalone/list.go))? It seems like this is where processes of daprd/dapr are found and returned by `dapr list` but the return list only contains processes of `daprd` that are linked to a corresponding `dapr` process. So if we ran a sidecar directly, there would be no parent `dapr` process and the `daprd` instance still wouldn't show up using `dapr list` correct? 

For now, I'll update the code to find both forms of `daprd` but if the end goal is to use `dapr list` to find running apps, I think we still wouldn't be able to get instances of `daprd` that were run directly.

src/services/processProvider.ts
```diff
@@ -18,13 +19,19 @@ export interface ProcessProvider {
...
-
         return processes
-            .filter(process => process.name === name)
+            .filter(process => process.name === name || this.hasDaprdPath(process))
             .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid }));
     }
+
+    hasDaprdPath(process: psList.ProcessDescriptor): boolean {
+        const daprdPath = which.sync('daprd', {nothrow: true});
+        const executable = daprdPath !== null ? daprdPath : which.sync('daprd.exe', {nothrow: true});
```
philliphoff, line 29: We shouldn't need to check for `daprd.exe` on Mac/Linux as that'd only be for Windows, right?

src/services/processProvider.ts
```diff
@@ -18,13 +19,19 @@ export interface ProcessProvider {
...
+        return executable !== null ? (process.cmd?.indexOf(executable) !== -1) : false;
```
philliphoff, line 30: This would check for the executable within the entire command line, right, which could pick up non-Dapr processes that just happen to refer to the executable path (e.g. `ls -l <executable>`).  I would think we'd only be concerned with commands that *start* with the executable path (with a trailing space, so as to not pick up `daprd` prefixed executables.

src/services/processProvider.ts
```diff
@@ -18,13 +19,19 @@ export interface ProcessProvider {
...
         const processes = await psList();
-
         return processes
-            .filter(process => process.name === name)
+            .filter(process => process.name === name || this.hasDaprdPath(process))
             .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid }));
     }
+
+    hasDaprdPath(process: psList.ProcessDescriptor): boolean {
+        const daprdPath = which.sync('daprd', {nothrow: true});
```
philliphoff, line 28: This would only pick up Dapr instances when `daprd` if it were in the `PATH`, right?  While we recommend adding it to `PATH`, it's not required.  The user can configure the extension to use a `daprd` executable from any arbitrary location.

src/services/processProvider.ts
sk593: Wouldn't only continuously running instances that contain the executable path be registered (which would be running applications)? The `ls` command and other commands that just reference the executable wouldn't stay running long enough for the extension to refresh and register them as `daprd` processes, right? I do agree though that this is something that should have been handled so I'll update the code to use `startWith()` instead of just checking whether it contains the string.

src/services/processProvider.ts
sk593: Yes, the which module requires `daprd` to be in the `PATH` environment. Is it possible to require users to add it to the `PATH` before using the extension? I'm not sure that any other way of seeing if the process starts with the `daprd` executable would work better. Since the `daprd` processes do all start with something like `<executable> --app-id...` it is possible to check that what comes before `--app-id` is a `daprd` executable. However, if the CLI code changes, then this might disrupt what happens with the process finder. 

I'm not sure how likely it is the CLI code would change but would this be a better way to find the processes?

src/services/processProvider.ts
```diff
@@ -12,27 +14,50 @@ export interface ProcessInfo {
...
+
 function getWmicValue(line: string): string {
     const index = line.indexOf('=');
 
     return line.substring(index + 1);
 }
 
 export class WindowsProcessProvider implements ProcessProvider {
-    async listProcesses(name: string): Promise<ProcessInfo[]> {
+    async listProcesses(name: string, _daprdPath: string): Promise<ProcessInfo[]> {
```
philliphoff, line 60: Nit: If an argument isn't used (and isn't required for positional reasons, it can simply be omitted).

src/services/processProvider.ts
```diff
@@ -12,27 +14,50 @@ export interface ProcessInfo {
...
         return processes
-            .filter(process => process.name === name)
+            .filter(process => process.name === name || this.hasDaprdPath(process, daprdPath))
             .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid }));
     }
+
+    hasDaprdPath(process: psList.ProcessDescriptor, daprdPath: string): boolean | undefined {
+        const daprd = which.sync('daprd', {nothrow: true});
+        if(daprd !== null) { //check if daprd is in PATH env
+            return process.cmd?.startsWith(daprd);
```
philliphoff, line 31: In both this `startsWith()` and the next, it will catch paths that just happen to have the path as a prefix (e.g. `/folder/dapr` and `/folder/daprabc` would both be caught).  I don't think we actually check whether the `daprd` is in the path; just looking for a file-like path is probably enough.

sk593: If we checked for the daprd path returned with a trailing whitespace (i.e. ".../daprd "), wouldn't that also fix the problem? Or when you just looking for a file-like path is enough, do you mean that the logic similar to the third if case is enough for all three scenarios?

src/services/processProvider.ts
```diff
@@ -12,19 +10,40 @@ export interface ProcessInfo {
...
+            return process.cmd?.startsWith(daprdPath.concat(" "))
+        } else { //if no config path, check if filepath is executable
+            const daprdEndPoint = "/daprd ";
+            const endpoint = process.cmd?.indexOf(daprdEndPoint);
+            if(endpoint !== undefined && endpoint !== -1) {
+                const executable = process.cmd?.substring(0, endpoint + daprdEndPoint.length - 1)
+                if(executable !== undefined) {
+                    try {
+                        fs.accessSync(executable, fs.constants.X_OK);
+                        return process.cmd?.startsWith(executable.concat(" "));
```
philliphoff, line 36: I'm not sure what this is testing; haven't it already been tested that the substring from the beginning of the command to the `/daprd ` is a valid file?  Should this just return `true`?

src/services/processProvider.ts
```diff
@@ -12,19 +10,40 @@ export interface ProcessInfo {
...
+    async listProcesses(name: string, daprdPath: string): Promise<ProcessInfo[]> {
+        const processes = await psList();
         return processes
-            .filter(process => process.name === name)
+            .filter(process => process.name === name || this.hasDaprdPath(process, daprdPath))
             .map(process => ({ name: process.name, cmd: process.cmd ?? '', pid: process.pid }));
     }
+
+    hasDaprdPath(process: psList.ProcessDescriptor, daprdPath: string): boolean | undefined {
+        if (daprdPath !== "daprd") { //check if config path provided
```
philliphoff, line 26: Nit: we've standardized on single-quotes for strings.

package.json
```diff
@@ -472,6 +472,7 @@
 		"@types/terser-webpack-plugin": "^5.0.2",
 		"@types/vscode": "^1.57.0",
 		"@types/webpack": "^4.41.26",
+		"@types/which": "^2.0.1",
```
philliphoff, line 475: This should no longer be required?

philliphoff (approved):
Just a couple of last comments, then it looks ok.

src/services/processProvider.ts
sk593: Yes, you're right. I'll update the function

Issue #167, opened 2021-05-13
Components panel in Explorer that shows active components being used in app


BigMorty:

jasonviviano:
## Description
After a Dapr application is expanded in the Application tree view, show a folder that contains all Components that the Dapr application uses. This folder can be expanded to reveal all the components used.

philliphoff:
Closing; enhancements to the current state will be tracked as separate items.

Issue #161, opened 2021-05-10
Expose more metadata in the Explorer about the services that make up the application


BigMorty:

jasonviviano:
## Description
Add a panel to the extension Explorer that displays a Dapr applications metadata.

### Metadata Wanted

•	Dapr HTTP Port 
•	Dapr gRPC Port 
•	Age 
•	Command 
•	App Port

philliphoff:
Closing; enhancements to the current state will be tracked as separate items.

PR #186, opened 2021-06-21
Added Components Metadata #161

sk593:
Added component metadata (name of component) to the application tree view. this addresses one aspect of issue #161 and addresses #167 that requests to expose more metadata. An example of the components view is shown below

Screen Shot 2021-06-22 at 11 02 34 AM

### DIFF
--- a/src/services/daprClient.ts
+++ b/src/services/daprClient.ts
@@ -14,6 +14,7 @@ export interface DaprClient {
     invokeGet(application: DaprApplication, method: string, token?: vscode.CancellationToken): Promise<unknown>;
     invokePost(application: DaprApplication, method: string, payload?: unknown, token?: vscode.CancellationToken): Promise<unknown>;
     publishMessage(application: DaprApplication, pubSubName: string, topic: string, payload?: unknown, token?: vscode.CancellationToken): Promise<void>;
+    getMetadata(application: DaprApplication, token?: vscode.CancellationToken): Promise<DaprMetadata>;
 }
 
 function manageResponse(response: HttpResponse): unknown {
@@ -60,4 +61,22 @@ export default class HttpDaprClient implements DaprClient {
 
         await this.httpClient.post(url, payload, { json: true }, token);
     }
+
+    async getMetadata(application: DaprApplication, token?: vscode.CancellationToken | undefined): Promise<DaprMetadata>  {
+        const originalUrl = `http://localhost:${application.httpPort}/v1.0/metadata`;
+
+        const response = await this.httpClient.get(originalUrl, { allowRedirects: false }, token);
+        
+        return manageResponse(response) as DaprMetadata;
+    }
 }
+
+export interface DaprMetadata {
+    components: DaprComponentMetadata[];
+  }
+  
+  export interface DaprComponentMetadata {
+      name: string;
+      type: string;
+      version: string;
+  }
--- a/src/views/applications/daprApplicationNode.ts
+++ b/src/views/applications/daprApplicationNode.ts
@@ -5,9 +5,10 @@ import * as vscode from 'vscode';
 import TreeNode from "../treeNode";
 import { DaprApplication } from '../../services/daprApplicationProvider';
 import DaprComponentsNode from "./daprComponentsNode";
+import { DaprClient } from '../../services/daprClient';
 
 export default class DaprApplicationNode implements TreeNode {
-    constructor(public readonly application: DaprApplication) {
+    constructor(public readonly application: DaprApplication, public readonly daprClient: DaprClient) {
     }
 
     getTreeItem(): Promise<vscode.TreeItem> {
@@ -21,6 +22,6 @@ export default class DaprApplicationNode implements TreeNode {
     }
 
     getChildren(): TreeNode[] {
-        return [new DaprComponentsNode('Components', this.application)];
+        return [new DaprComponentsNode(this.application, this.daprClient)];
     }
 }
--- a/src/views/applications/daprApplicationNode.ts
+++ b/src/views/applications/daprApplicationNode.ts
@@ -4,18 +4,23 @@
 import * as vscode from 'vscode';
 import TreeNode from "../treeNode";
 import { DaprApplication } from '../../services/daprApplicationProvider';
+import DaprComponentsNode from "./daprComponentsNode";
 
 export default class DaprApplicationNode implements TreeNode {
     constructor(public readonly application: DaprApplication) {
     }
 
     getTreeItem(): Promise<vscode.TreeItem> {
-        const item = new vscode.TreeItem(this.application.appId);
+        const item = new vscode.TreeItem(this.application.appId, vscode.TreeItemCollapsibleState.Collapsed);
 
         item.contextValue = 'application';
 
         item.iconPath = new vscode.ThemeIcon('globe');
 
         return Promise.resolve(item);
+    }
+
+    getChildren(): TreeNode[] {
+        return [new DaprComponentsNode('Components', this.application)];
     }
 }
--- a/src/views/applications/daprComponentsNode.ts
+++ b/src/views/applications/daprComponentsNode.ts
@@ -2,19 +2,24 @@
 // Licensed under the MIT license.
 
 import * as vscode from 'vscode';
+import * as nls from 'vscode-nls';
 import TreeNode from "../treeNode";
 import { DaprApplication } from '../../services/daprApplicationProvider';
 import DaprMetadataNode from './daprMetadataNode';
-import AxiosHttpClient from '../../services/httpClient';
-import HttpDaprClient, { DaprMetadata } from '../../services/daprClient';
+import { DaprClient, DaprMetadata } from '../../services/daprClient';
+import { getLocalizationPathForFile } from '../../util/localization';
+
+const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
 
 
 export default class DaprComponentsNode implements TreeNode {
-    constructor(public readonly metadata: string, public readonly application: DaprApplication) {
+    constructor(public readonly application: DaprApplication, public readonly daprClient: DaprClient) {
     }
 
     getTreeItem(): Promise<vscode.TreeItem> {
-        const item = new vscode.TreeItem(this.metadata, vscode.TreeItemCollapsibleState.Collapsed);
+        const label = localize('views.applications.daprComponentsNode.componentNode', 'Components');
+
+        const item = new vscode.TreeItem(label, vscode.TreeItemCollapsibleState.Collapsed);
 
         item.contextValue = 'components';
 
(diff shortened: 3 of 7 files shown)


philliphoff:
@sk593 Can you add a screenshot showing the new UX; that often helps with reviews.

src/services/daprClient.ts
```diff
@@ -60,4 +61,12 @@ export default class HttpDaprClient implements DaprClient {
 
         await this.httpClient.post(url, payload, { json: true }, token);
     }
+
+    async getMetadata(application: DaprApplication, token?: vscode.CancellationToken | undefined): Promise<unknown>  {
```
philliphoff, line 65: Does the structure of the data returned from this endpoint change based on context, or is it always the same?  If the former, can we statically type the return result rather than return `unknown`?  (The other calls use `unknown` because their return values depend on what's returned by the application and so can be literally anything, but this is a call to the sidecar itself.)

src/views/applications/daprComponentsNode.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private async getMetadata(application: DaprApplication, token?: vscode.CancellationToken | undefined): Promise<unknown>  {
+        const daprClient = new HttpDaprClient(new AxiosHttpClient());
+        return await daprClient.getMetadata(application, token);
+    }
+    
+}
+
+export interface MyResponse {
```
philliphoff, line 39: Nit: naming

src/views/applications/daprComponentsNode.ts
```diff
@@ -0,0 +1,50 @@
...
+    "id": string,
+    "actors": JSON[],
+    "extended": string[],
+    "components": JSON[]
```
philliphoff, line 43: Nit: use JS rather than JSON syntax for defining interfaces. If there are properties returned we don't use, either fully describe them (for the purposes of documentation and future use) or omit them. So, at a minimum:

```typescript
export interface DaprMetadata {
  components: Components[];
}

export interface DaprComponentMetadata {
    name: string;
    type: string;
    version: string;
}
```

src/views/applications/daprMetaDataNode.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as vscode from 'vscode';
+import TreeNode from "../treeNode";
+
+
+
+export default class DaprMetaDataNode implements TreeNode {
```
philliphoff, line 9: Suggestion: `DaprMetadataNode` (metadata as one word instead of two).

src/views/applications/daprMetaDataNode.ts
```diff
@@ -0,0 +1,24 @@
...
+    constructor(public readonly metadata: string) {
+    }
+
+    getTreeItem(): Promise<vscode.TreeItem> {
+        const item = new vscode.TreeItem(this.metadata);
+
+        item.contextValue = 'application';
```
philliphoff, line 16: The `contextValue` should probably be `metadata` or similar.

src/views/applications/daprComponentsNode.ts
```diff
@@ -0,0 +1,50 @@
...
+
+
+export default class DaprComponentsNode implements TreeNode {
+    constructor(public readonly metadata: string, public readonly application: DaprApplication) {
+    }
+
+    getTreeItem(): Promise<vscode.TreeItem> {
+        const item = new vscode.TreeItem(this.metadata, vscode.TreeItemCollapsibleState.Collapsed);
+
+        item.contextValue = 'application';
```
philliphoff, line 19: The `contextValue` should probably be `components` or similar.

src/services/daprClient.ts
sk593: No, the dapr Metadata API returns the same data fields about a service every time so I can update the code to return a static type

sk593:
I initially pushed changes to address the comments but it caused some Linux package checks to fail. I accidentally uncommitted those changes when trying to fix the error so the original PR doesn't have the initial commits I made. But the last commit should have all the original changes, as well as the requested updates @philliphoff

src/views/applications/daprApplicationNode.ts
```diff
@@ -4,18 +4,23 @@
...
 
         item.contextValue = 'application';
 
         item.iconPath = new vscode.ThemeIcon('globe');
 
         return Promise.resolve(item);
     }
+
+    getChildren(): TreeNode[] {
+        return [new DaprComponentsNode('Components', this.application)];
```
philliphoff, line 24: We should make `'Components'` localizable.  See how we use the `localize()` function elsewhere.

src/views/applications/daprMetadataNode.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as vscode from 'vscode';
+import TreeNode from "../treeNode";
+
+export default class DaprMetadataNode implements TreeNode {
```
philliphoff, line 7: Should this be `DaprComponentNode` and accept a `name` constructor parameter?  I'm not sure if we'd ever end up re-using this node for *other* kinds of metadata, so it might make sense to more closely tie to its actual usage.

src/views/applications/daprComponentsNode.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    async getChildren(): Promise<TreeNode[]> {
+        const responseData = await this.getMetadata(this.application);
+        const components = responseData.components;
+        return components.map(comp => new DaprMetadataNode(comp.name));
+    }
+
+    private async getMetadata(application: DaprApplication, token?: vscode.CancellationToken | undefined): Promise<DaprMetadata>  {
+        const daprClient = new HttpDaprClient(new AxiosHttpClient());
```
philliphoff, line 33: We should pass in an existing `DaprClient` through the constructor rather than creating one where (which may mean passing in a `DaprClient` to the tree provider (so it can be passed down).

src/views/applications/daprComponentsNode.ts
```diff
@@ -0,0 +1,38 @@
...
+        const item = new vscode.TreeItem(this.metadata, vscode.TreeItemCollapsibleState.Collapsed);
+
+        item.contextValue = 'components';
+
+        item.iconPath = new vscode.ThemeIcon('archive');
+
+        return Promise.resolve(item);
+    }
+
+    async getChildren(): Promise<TreeNode[]> {
```
philliphoff, line 26: How do we want to deal with the (admittedly edge case) of an application having *no* components.  That would leave a dangling `Components` node.  That *might* be ok (maybe with adding a `'0 items'` sub-label), but we might also decide to omit the `Components` node altogether in that case, or perhaps have a "none" child node.

That may be a good question to bring up to the PMs in the next standup.

src/views/applications/daprMetadataNode.ts
sk593: I was thinking that it could be reused for the service metadata feature (i.e. age of app, app id, app port, dapr port, etc) since it takes in a generic metadata parameter. But we could have separate nodes for the different types of metadata that we want to expose in the tree view.

src/views/applications/daprComponentsNode.ts
sk593: We talked about being able to add components as a next-step feature for the application view. The Components node would need to be there for that user scenario, but we can definitely add a sub-label to denote that there are no components until that feature is implemented. I'll bring it up in today's meeting, though, and see what the PMs say

philliphoff: I think that's ok for now, though as soon as we want to add component-specific commands (e.g. open component YAML), we'll likely want to create a specialized class.

src/views/applications/daprComponentsNode.ts
```diff
@@ -2,19 +2,24 @@
...
-import HttpDaprClient, { DaprMetadata } from '../../services/daprClient';
+import { DaprClient, DaprMetadata } from '../../services/daprClient';
+import { getLocalizationPathForFile } from '../../util/localization';
+
+const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
 
 
 export default class DaprComponentsNode implements TreeNode {
-    constructor(public readonly metadata: string, public readonly application: DaprApplication) {
+    constructor(public readonly application: DaprApplication, public readonly daprClient: DaprClient) {
```
philliphoff, line 16: Nit: unless they have to be accessible outside of the class, use `private readonly`.

src/views/applications/daprComponentsNode.ts
```diff
@@ -24,14 +29,18 @@ export default class DaprComponentsNode implements TreeNode {
...
         const responseData = await this.getMetadata(this.application);
         const components = responseData.components;
-        return components.map(comp => new DaprMetadataNode(comp.name));
+        if(components.length > 0) {
+            return components.map(comp => new DaprMetadataNode(comp.name, 'database'));
+        }
+        return [new DaprMetadataNode(label, 'warning')];
     }
 
     private async getMetadata(application: DaprApplication, token?: vscode.CancellationToken | undefined): Promise<DaprMetadata>  {
```
philliphoff, line 41: Nit: as this is now only a single line function, used from a single location, is it necessary?

src/views/applications/daprMetadataNode.ts
philliphoff: I think that's ok for now, though as soon as we want to add component-specific commands (e.g. open component YAML), we'll likely want to create a specialized class.

src/views/applications/daprComponentsNode.ts
philliphoff: A subnode seems fine for now, though we might still consider a `(n items)` sub-label in the Components node.

src/views/applications/daprMetadataNode.ts
```diff
@@ -5,15 +5,15 @@ import * as vscode from 'vscode';
 import TreeNode from "../treeNode";
 
 export default class DaprMetadataNode implements TreeNode {
-    constructor(public readonly metadata: string) {
+    constructor(public readonly metadata: string, public readonly icon: string) {
```
philliphoff, line 8: Perhaps `themeIconId: string` to reinforce its usage (as opposed to a resource path string, which is another way to specify an icon)?

philliphoff:
@sk593 A couple of final nits, but otherwise looks good.  Let's also make sure there is an issue filed (maybe assigned to @AaronCrawfis) to get a better set of icons into the product at some point.

src/views/applications/daprComponentsNode.ts
```diff
@@ -6,14 +6,14 @@ import * as nls from 'vscode-nls';
...
-import { DaprClient, DaprMetadata } from '../../services/daprClient';
+import { DaprClient } from '../../services/daprClient';
 import { getLocalizationPathForFile } from '../../util/localization';
 
 const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
 
 
 export default class DaprComponentsNode implements TreeNode {
-    constructor(public readonly application: DaprApplication, public readonly daprClient: DaprClient) {
+    constructor(private readonly application: DaprApplication, private readonly daprClient: DaprClient, private readonly token?: vscode.CancellationToken | undefined) {
```
philliphoff, line 16: I'd simply omit the `token`; whatever cancellation it's supposed to represent is likely very different from the notion of returning children of the tree.  If we were being super sophisticated we'd probably create a token in `getChildren()` that represents, say, a timeout of 60s and pass that to the `getMetadata()` function.  However, given the call is unlikely to ever timeout (i.e. it's far more likely to either immediately fail or nearly immediately succeed), it's generally not worth it.

philliphoff (approved):

PR #188, opened 2021-07-14
Added application details views 

sk593:
Added applications details view to expose metadata as outlined in issue #161. When an application name is clicked on, the app details pane populates the data 

UI:
Screen Shot 2021-07-14 at 10 24 30 AM


src/services/daprApplicationProvider.ts
```diff
@@ -39,6 +41,34 @@ function getHttpPort(cmd: string): number {
...
+    const portRegEx = /--app-port "?(?<port>\d+)"?/g;
+        
+    const portMatch = portRegEx.exec(cmd);
+    
+    const portString = portMatch?.groups?.['port'];
+    
+    if (portString !== undefined) {
+        return parseInt(portString, 10);
+    } else {
+        return 4500;
```
philliphoff, line 68: Is this really the default when no application port is specified?  Should we just return `undefined` and then handle that case in the tree view (e.g. show `<None>` or something)?

package.json
```diff
@@ -72,6 +73,11 @@
 				"title": "%vscode-dapr.applications.publish-message.title%",
 				"category": "Dapr"
 			},
+			{
```
philliphoff, line 76: Nit: to avoid confusion, perhaps choose a command ID that's different than the view ID.

src/views/details/appDetailsTreeDataProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+    getChildren(): TreeNode[] {
+        const labels = ["App ID", "App Port", "Dapr HTTP Port", "Dapr GRPC Port", "Process ID"];
+        const list : DaprAppDetailsNode[] = []
+        for(let i = 0; i < labels.length; i++) {
+            list.push(new DaprAppDetailsNode(labels[i], this.appDetails[i]));
+        }
+        return list;
+    }
+
+    getAppDetails(application: DaprApplication | undefined) : void {
```
philliphoff, line 44: Should this be `setAppDetails()`?

src/views/details/appDetailsTreeDataProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+        if(application !== undefined) {
+            this.appDetails = [application?.appId.toString(), application?.appPort.toString(), application?.httpPort.toString(), application?.grpcPort.toString(), application?.pid.toString()]
```
philliphoff, line 46: After being set, does this need to fire the changed event so VS Code will refresh the tree?

src/services/daprApplicationProvider.ts
sk593: `dapr list` uses 0 when there's no port specified. Would that be good? Otherwise, yes I agree that a `None` description would be better.

src/views/details/appDetailsTreeDataProvider.ts
sk593: VS Code updates the tree as the data in the tree changes. So choosing a new application node updates the tree data

src/services/daprApplicationProvider.ts
philliphoff: I would still go with something more user-friendly (e.g. `<None>`).

src/views/applications/daprComponentsNode.ts
```diff
@@ -29,13 +30,14 @@ export default class DaprComponentsNode implements TreeNode {
     }
 
     async getChildren(): Promise<TreeNode[]> {
+        //TO DO: FIX WARNING NODE
```
philliphoff, line 33: What needs to be fixed?

src/views/details/detailsTreeDataProvider.ts
```diff
@@ -0,0 +1,61 @@
...
+    getChildren(): TreeNode[] {
+        const list : DaprDetailsNode[] = []
+        for(let i = 0; i < this.detailLabels.length; i++) {
+            list.push(new DaprDetailsNode(this.detailLabels[i], this.details[i]));
+        }
+        return list;
+    }
+
+    setAppDetails(application: DaprApplication | undefined) : void {
+        this.detailLabels = ["App ID", "App Port", "Dapr HTTP Port", "Dapr GRPC Port", "Dapr Process ID"];
```
philliphoff, line 46: These (and below) labels should be localizable.

src/views/details/detailsTreeDataProvider.ts
```diff
@@ -0,0 +1,61 @@
...
+        if(application !== undefined) {
+            this.details = [application?.appId.toString(), 
+                application?.appPort !== undefined ? application?.appPort.toString() : "None", 
+                application?.httpPort.toString(), application?.grpcPort.toString(), application?.pid.toString()]
+        }
+    }
+
+    setComponentDetails(component: DaprComponentMetadata | undefined) : void {
+        this.detailLabels = ["Name", "Type", "Version"]
```
philliphoff, line 55: This is *ok*, but can lead to fragility in that the order can be easily mixed up.  Perhaps something like:

```typescript
this.details = [
  { label: 'Name', value: component.name },
  { label: 'Type', value: component.type },
  ...
];
```

philliphoff (approved):
Just some final comments; once at least the localization ones are addressed, this should be good to go.

src/views/applications/daprComponentsNode.ts
sk593: My bad, it was a comment to myself but I forgot to remove it before committing.

PR #190, opened 2021-08-02
Continuing Dashboard Issue #159

sk593:
Picked up the existing changes from PR #180


src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+           throw new Error(localize('dashboard.findOpenPort.noOpenPort', 'No open port found.'))
+        });
+    }
+
+    processSpawn() : void{
+        const portNumber = this.getPortUsed();
+        if(portNumber !== undefined) {
+            void new Promise((_resolve, reject) => {
+                void Process.exec(`${this.daprPathProvider()} dashboard -p ${portNumber}`)
+                setTimeout(reject, 10 * 1000) //timeout after dashboard command is run
```
philliphoff, line 52: This would seem to *always* reject (after 10 seconds)?  `Process.exec()` also resolves only once the command completes, but this is a long-running process, so it would be expected to never complete (unless specifically killed), right?

The simplest approach to me would be:

 - If no process yet spawned:
   1. Find a free port.
   1. Spawn the dashboard using the free port.
   1. If successful, store both the `Process` and the port used.
   1. Launch the browser using the stored port.
 - Otherwise:
   1. Launch the browser using the stored port.

This ensures we always find a new free port when starting the dashboard, but also allows us to kill the process when the extension is shutdown.

src/commands/openDaprDashboard.ts
```diff
@@ -0,0 +1,14 @@
...
+import ClassBasedDaprDashboardProvider from '../services/daprDashboardProvider';
+import { UserInput } from '../services/userInput';
+
+
+export async function openDaprDashboard(ui: UserInput, classBasedDaprDashboardProvider: ClassBasedDaprDashboardProvider): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    await vscode.env.openExternal(vscode.Uri.parse(await classBasedDaprDashboardProvider.startDashboard()));
+}
+
+const createOpenDaprDashboardCommand = (ui: UserInput, classBasedDaprDashboardProvider: ClassBasedDaprDashboardProvider) => (): Promise<void> => openDaprDashboard(ui, classBasedDaprDashboardProvider);
```
philliphoff, line 13: The command should accept a `DaprDashboardProvider` (i.e. the interface) rather than a specific implementation.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    async startDashboard(): Promise<string> {
+        if(this.port !== undefined) {
+            return `http://localhost:${this.port}`
+        }
+        else {
+            await this.spawnDashboardInstance();
+
+            if(this.openPort !== undefined) {
+                await tcp.waitForStatus(this.openPort, 'localhost', true, 500, 4000).then(() => {
```
philliphoff, line 31: If `async`/`await` are being used, there's generally no reason to *also* use `then()` and `catch()`. I think it would help make the flow more straightforward to remove them.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    async findOpenPort(): Promise<number> {
+        return await portfinder.getPortPromise().catch(() => {
+            throw new Error(localize('dashboard.findOpenPort.noOpenPort', 'No open port found.'))
+        });
+    }
+
+    async processSpawn() : Promise<void> {
+        this.openPort = await this.findOpenPort();
+        if(this.openPort !== undefined) {
```
philliphoff, line 56: Nit: `findOpenPort()` should only return a port (or throw), so checking for undefined shouldn't be necessary.  (Removal of the `then()` and `catch()` may help make that more clear.)

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+                await tcp.waitForStatus(this.openPort, 'localhost', true, 500, 4000).then(() => {
+                    this.port = this.openPort; //store port value if localhost port connection is successful
+                }, (err) => {
+                    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+                    const msg = err + localize('dashboard.startDashboard.statusTimeout', ': unable to connect to localhost port ') + this.openPort?.toString();
+                    throw new Error(msg)
+                });
+            }
+            // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+            return `http://localhost:${this.port}`
```
philliphoff, line 40: Rather than duplicate constructing the URL, perhaps check to see if spawning is necessary (and do so) first, then just have a single return with the URL.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    async spawnDashboardInstance(): Promise<void>{
```
philliphoff, line 44: Nit: it seems like this function could be removed in favor of just calling `processSpawn()` directly, or perhaps just pulling its implementation here.

src/services/daprDashboardProvider.ts
sk593: I think the `.then()` is needed since we only want to set the value of the port if a connection to the localhost site is actually established and not otherwise. I tried removing `await` but there is some command that is lagging (either when running the dashboard process or when trying to set the port value on a successful connection) that stops the browser from opening when the button is clicked on the first time. Without both `await` and the `then()`, the browser doesn't open immediately and the user has to click the button multiple times

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+        void Process.exec(`${this.daprPathProvider()} dashboard -p ${this.openPort}`)
+            .catch(() => {
+                throw new Error(localize('dashboard.spawnDashboard.spawnFailure', 'Dashboard process failed to spawn'))
+            })
+        
+        return;
+    }
+
+    async findOpenPort(): Promise<number> {
+        return await portfinder.getPortPromise().catch(() => {
```
philliphoff, line 53: I don't think there's any reason for the `catch()` here (since we're using `async`/`await`); if we're just adjusting the error message then we should use normal `try`/`catch` blocks.  If we don't really need to adjust the message then we can probably just call `getPortPromise()` directly and not have `findOpenPort()` at all.

src/services/daprDashboardProvider.ts
philliphoff: A problem I see here is that the user can call the command twice (the first spawns the dashboard) if the port takes a while to start (because the `openPort` hasn't been set).  Another approach is to have a field in the class that stores a `Promise` that reflects both an attempt to start the dashboard and a wait for it to be ready.  If the field hasn't been set, that means it's the first call, so assign the promise (which starts the startup process) and wait for it to resolve.  If the field *has* been set, that means a subsequent call made before the dashboard is ready, and we just have to wait for it to be done.

```typescript
private startup: Promise<number>;

async startDashboard(): Promise<string> {
  if (!this.startup) {
    this.startup = new Promise(
      (resolve, reject) => {
        // Find open port...
        // Spawn dashboard...
        // Wait for dashboard to be ready...
      });
  }
  
  const openPort = await this.startup;

 return `http://localhost:${port}`;
}
```

sk593: I condensed the logic so that it is part of one Promise call rather than splitting them up. I couldn't avoid using a try/catch block since certain errors I tested would return before a `reject()` call could be made. Let me know if those changes better align with what you were saying about using a single Promise to handle everything.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        if (!localStartup) {
+            localStartup = new Promise(
+                // eslint-disable-next-line @typescript-eslint/no-misused-promises, no-async-promise-executor
+                async (resolve, reject) => {
+                    try {
+                        const openPort = await portfinder.getPortPromise();
+                        void Process.exec(`${this.daprPathProvider()} dashboard -p ${openPort}`)
+                        
+                        const successfulPort = await tcp.waitForStatus(openPort, 'localhost', true, 500, 4000).then(() => {
```
philliphoff, line 34: This `then()` shouldn't be needed, right?  We already know the port--we're just waiting for the status to change.

src/services/daprDashboardProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+                            return openPort; 
+                        });
+                        
+                        resolve(successfulPort);
+                    } catch(error) {
+                        this.startUp = undefined;
+                        // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+                        const msg = localize('dashboard.startDashboard.startupError', 'Dashboard instance failed to start. ') + error;
```
philliphoff, line 42: Nit: use template strings to avoid linter warning (e.g. `` `Dashboard instance failed to start: ${error}` ``)

sk593: Localize only takes a string literal so it wouldn't allow template strings without another linter warning.

philliphoff: Ah, right, though I think it allows use of template-like tokens and then passing each value to be substituted as an additional argument.

philliphoff (approved):

.devcontainer/library-scripts/azcli-debian.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
philliphoff, line 1: Is this a real change?

.vscode/tasks.json
```diff
@@ -23,6 +23,18 @@
 			"presentation": {
 				"reveal": "never"
 			}
+		},
```
philliphoff, line 26: Is this a real change?

sk593: This and the other file you commented on were part of a commit pushed by Jason (I'm assuming by accident). Should I just delete them? I don't think they're supposed to be there

philliphoff: They don't appear to be actively part of the change, so I'd just remove them.

PR #200, opened 2021-08-11
Fix process provider offsets.

### LINKED ISSUE #199: Dapr explorer missing applications on Windows
On Windows, the Dapr explorer shows up to two running Dapr applications, even if more are running.

The issue is in the Windows-specific process provider, which miscalculates the offsets of items returned by the WMIC command, and causes it to get far enough out of sync that only two items are actually shown in the explorer.

philliphoff:
Fixes up the item offsets used by the process provider due the addition of the new property.

Resolves #199.


pratiksanglikar (approved):
Apart for the comment, looks good to me!

philliphoff:
This work is supplanted by #206 which completely swaps the implementation.

PR #222, opened 2022-01-10
Extract default ASP.NET port from launch setting profile

### LINKED ISSUE #214: Fail to invoke the application method for a .NET 6 Web API project
**OS:** Win10
**Build Version:** 20211202.11
**.NET Version:** 6.0.101
**Regression:** Not a Regression

**Repro Steps:**
1. Create a new .NET 6.0 ASP.NET Web API application by "dotnet new webapi --no-https --name netwebapp".
2. Open the application in VS Code -> Select "Dapr: Scaffold Dapr Tasks" from Command Palette -> Select ".NET Core Launch (web)" when asked to select a configuration -> Enter an application ID (Select the default "app") -> Enter an application port (Select the default "5000").
3. Debug with configuration ".NET Core Launch (web) with Dapr".
4. F1 -> Dapr: Invoke (GET) Application Method -> Select a Dapr application "app" -> Enter the application method "WeatherForecast" to invoke.
5. Check whether the method is invoked or not.

**Expect:**
The method is invoked without any error.

**Actual:**
Fail to invoke the application method with an error "Request failed with status code 500".
image

**More Info:**
1. This issue also reproduces on the previous released dapr extension.
2. This issue does not reproduce for Node.js project.
3. This issue does not reproduce for .NET 5 Web API project.

philliphoff:
Starting with .NET 6, new ASP.NET applications are scaffolded to use a dynamically generated port, rather than `5000` used by earlier frameworks.  This change updates our default port logic to sniff the `launchSettings.json`, extracting the application URL for the "project" profile (i.e. the profile used by the Kestrel host, which is also now the default for ASP.NET applications), and using its port number as the default provided to the user.  (The user can, of course, overwrite that value with their own.)

Resolves #214.

### DIFF
--- a/src/commands/scaffoldDaprTasks.ts
+++ b/src/commands/scaffoldDaprTasks.ts
@@ -58,15 +58,19 @@ async function getDefaultDotnetPort(folder: vscode.WorkspaceFolder | undefined):
                 const projectProfile = Object.values(launchSettingsJson.profiles).find(profile => profile.commandName === 'Project');
 
                 if (projectProfile?.applicationUrl) {
-                    try {
-                        const applicationUrl = new URL(projectProfile.applicationUrl);
-    
-                        if (applicationUrl.port) {
-                            return parseInt(applicationUrl.port, 10);
+                    const applicationUrls = projectProfile.applicationUrl.split(';');
+
+                    for (const applicationUrl of applicationUrls) {
+                        try {
+                            const url = new URL(applicationUrl);
+                            
+                            if (url.protocol === 'http' && url.port) {
+                                return parseInt(url.port, 10);
+                            }
                         }
-                    }
-                    catch {
-                        // NOTE: Ignore any errors parsing the URL.
+                        catch {
+                            // NOTE: Ignore any errors parsing the URL.
+                        }
                     }
                 }
             }
--- a/src/commands/scaffoldDaprTasks.ts
+++ b/src/commands/scaffoldDaprTasks.ts
@@ -3,6 +3,7 @@
 
 import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
+import { URL } from 'url';
 import { DaprTaskDefinition } from "../tasks/daprCommandTaskProvider";
 import { DaprdDownTaskDefinition } from "../tasks/daprdDownTaskProvider";
 import { getWorkspaceConfigurations, DebugConfiguration } from '../scaffolding/configurationScaffolder';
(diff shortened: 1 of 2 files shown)


src/commands/scaffoldDaprTasks.ts
```diff
@@ -34,40 +36,80 @@ const JavaPort = 8080;
...
+interface DotNetLaunchSettings {
+    profiles?: { [key: string]: DotNetProfile };
+}
+
+async function getDefaultDotnetPort(folder: vscode.WorkspaceFolder | undefined): Promise<number> {
+    if (folder) {
+        const launchSettingsFiles = await vscode.workspace.findFiles(new vscode.RelativePattern(folder, "**/Properties/launchSettings.json"));
+
+        if (launchSettingsFiles.length > 0) {
+            const launchSettingsBuffer = await vscode.workspace.fs.readFile(launchSettingsFiles[0]);
```
ravipal, line 53: When multiple launchSettings.json are found why limit to just the first one? Can we iterate all until a profile with project name found?

src/commands/scaffoldDaprTasks.ts
```diff
@@ -34,40 +36,80 @@ const JavaPort = 8080;
...
+            const launchSettingsContents = new TextDecoder('utf-8').decode(launchSettingsBuffer);
+            const launchSettingsJson = JSON.parse(launchSettingsContents) as DotNetLaunchSettings;
+
+            if (launchSettingsJson.profiles) {
+                const projectProfile = Object.values(launchSettingsJson.profiles).find(profile => profile.commandName === 'Project');
+
+                if (projectProfile?.applicationUrl) {
+                    try {
+                        const applicationUrl = new URL(projectProfile.applicationUrl);
```
ravipal, line 62: `projectProfile.applicationUrl` could have multiple urls (http, https) separated by ;.

ravipal (changes requested):

src/commands/scaffoldDaprTasks.ts
philliphoff: That's good to know; I'll update it to find the port of the first HTTP URL.  (By default, Dapr expects to use HTTP between the application and sidecar; HTTPS *can* be used, but it involves more configuration that would complicate the UX and is probably not needed at this point.)  We're definitely optimizing for the single project, single endpoint scenario when choosing the default port.

src/commands/scaffoldDaprTasks.ts
philliphoff: I used the first one as that represents the simplest case we support (a single project repo).  To properly support multi-project repos, additional UX is required (e.g. finding all projects and then asking the user which is the one being scaffolded).  Unfortunately, it's not totally deterministic to identify the project from the launch configuration (as it only has the build output path, which doesn't necessarily correspond to project paths).

I've added an issue to improve multi-project repo support when scaffolding (#223).

philliphoff: Updated to look at all launch settings files until a suitable profile/port was found.

src/commands/scaffoldDaprTasks.ts
philliphoff: Updated to handle multiple application URLs.

ravipal (approved):

PR #230, opened 2022-02-04
Remove Gulp as a task runner

### LINKED ISSUE #229: Move away from Gulp
The nagging NPM security alerts for the repo all come from our dependency on Gulp for building. As much as I love it compared to the alternatives, we should consider moving away from it in order not to cloud more meaningful alerts, should they occur.

philliphoff:
Removes Gulp as a task runner in favor of npm scripts.  As Gulp is essentially "done", it's not receiving updates which means it's a continual source of corporate governance alerts, irrelevant as they may be.

> NOTE: As the localization pipeline of `vscode-nls-dev` is Gulp-based, we lose that ability for now, but technically we weren't really using it anyway.  (The source is still prepped for localization, but we just won't do the transformation/inclusion of the localization files, which we don't have anyway.) We can revisit this if and when we get to 1.0 (general availability).

Resolves #229.

@bwateratmsft I hope your happy now.  :-)


bwateratmsft:
> @bwateratmsft I hope your happy now. :-)

😈

gulpfile.ts
```diff
@@ -110,14 +110,6 @@ function addI18nTask() {
         .pipe(gulp.dest('.'));
```
bwateratmsft, line 110: Any reason to keep this file anymore?

bwateratmsft (approved):

gulpfile.ts
philliphoff: My subconcious just wouldn't let me let it go.  Removed.  <sigh> :-)

bwateratmsft: image

_Phil letting Gulp go_

philliphoff: Gulp will make a comeback.  I just *know* it!

PR #238, opened 2022-06-03
Pull dependencies forward

philliphoff:
Updates dependencies to their latest versions.  In particular, uses the latest Azure Extension "Utils" package that replaces the `vscode-azureextensionui` package as well as eliminates the need for `fs-extra`.


src/commands/scaffoldDaprComponents.ts
```diff
@@ -22,7 +22,7 @@ async function scaffoldDaprComponents(scaffolder: Scaffolder, templateScaffolder
 
     const componentsPath = path.join(rootWorkspaceFolderPath, 'components');
 
-    await fse.ensureDir(componentsPath);
+    await fs.mkdir(componentsPath, { recursive: true });
```
ravipal, line 25: I assume this wouldn't throw when the directory is already present.

ravipal (approved):

src/commands/scaffoldDaprComponents.ts
philliphoff: Yes, when `recursive` is `true` it does not throw.

Issue #212, opened 2021-12-13
daprd-down postDebugTask fails to kill daprd process


asalvi0:
Does this occur consistently? No
Repro steps:
1. Create **launch.json** as follows:
`{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "go",
            "request": "launch",
            "name": "Go with Dapr",
            "mode": "debug",
            "program": "${workspaceFolder}\\cmd\\hnhandler",
            "console": "integratedTerminal",
            "preLaunchTask": "dapr-debug",
            "postDebugTask": "daprd-down",
        }
    ]
}`

1. Create **tasks.json** as follows:
 `{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "dapr",
            "label": "dapr-debug",
            "appId": "hnhandler",
            "appProtocol": "grpc",
            "appPort": 5000,
            "componentsPath": "dapr/components",
            "presentation": {
                "showReuseMessage": false,
                "clear": true,
                "close": true,
                "focus": true
           }
        },
        {
            "type": "daprd-down",
            "label": "daprd-down",
            "appId": "hnhandler",
            "presentation": {
                "showReuseMessage": false,
                "clear": true,
                "close": true
           }
        }
    ]
}`

2. Stop the debugger and the **dapr** and **daprd** processes will continue running. An error message is displayed in VSCode

**Action: vscode-dapr.tasks.dapr-down
Error type: ESRCH
Error Message: kill ESRCH**

Version: 0.5.0
OS: win32
OS Release: 10.0.19043
Product: Visual Studio Code
Product Version: 1.63.0
Language: en

<details>
<summary>Call Stack</summary>

```
process.kill per_thread.js:200:13
extension.js:2:1496054extension.js:2:1496054
DaprdDownTaskProvider.<anonymous> extension.js:2:1496034
fulfilled extension.js:2:1495316
processTicksAndRejections task_queues.js:93:5
```

</details>

v-mengwe:
**OS:** All
**Build Version:** 20220108.1

**Repro Steps:**

1. Create a new .NET 6.0 ASP.NET Web API application by "dotnet new webapi --no-https --name netwebapp".
2. Open the application in VS Code -> Select "Dapr: Scaffold Dapr Tasks" from Command Palette -> Select ".NET Core Launch (web)" when asked to select a configuration -> Enter an application ID (Select the default "app") -> Enter an application port (Select the default "5000").
[...]

PR #256, opened 2022-12-14
chore: change shut down daprd message

yukun-dong:
The log message here should tell user that the daprd shut down is complete, since the application process is killed in L35.  The original log "Shutting down daprd.." means it is still in progress.

### DIFF
--- a/src/tasks/daprdDownTaskProvider.ts
+++ b/src/tasks/daprdDownTaskProvider.ts
@@ -34,7 +34,7 @@ export default class DaprdDownTaskProvider extends CustomExecutionTaskProvider {
                             .filter(application => application.appId === daprdDownDefinition.appId)
                             .forEach(application => process.kill(application.pid, 'SIGKILL'));
                         
-                        writer.writeLine(localize('tasks.daprdDownTaskProvider.shutdownMessage', 'Shutting down daprd...'));
+                        writer.writeLine(localize('tasks.daprdDownTaskProvider.shutdownMessage', 'daprd shut down complete'));
                     });
             },
             /* isBackgroundTask: */ false,


philliphoff:
@yukun-dong If I recall correctly, it's phrased that way because we're only *requesting* that the process be terminated but do not know whether/when it's actually been done.  `SIGKILL` *should* cause an immediate termination (at least on unix-y machines), but it's not necessarily instantaneous.

It might be better to poll several times to ensure that the processes do actually, eventually, go down.

yukun-dong:
> @yukun-dong If I recall correctly, it's phrased that way because we're only _requesting_ that the process be terminated but do not know whether/when it's actually been done. `SIGKILL` _should_ cause an immediate termination (at least on unix-y machines), but it's not necessarily instantaneous.
> 
> It might be better to poll several times to ensure that the processes do actually, eventually, go down.

@philliphoff Thanks for you explanation! I checked the doc and found out `process.kill()` is just a signal sender rather than actually killing the process, so you are right.

When I stopped debugging using this extension and saw the following log, I thought it's still in progress and I waited a while for it, but actually it has ended so I think it is a little confusing. I expect the log here can tell the user when this killing process is really complete.

image

I'll see if I can make the log here better later, before that I'll turn this PR into a draft first.

yukun-dong:
> @yukun-dong In #212 I mention that we should probably switch to using `dapr stop` as a means to stopping a given Dapr sidecar rather than killing the process outright. That, at least, shifts the burden towards the CLI as it should know best how to do that. That said, it looks like the `dapr` CLI also runs `kill` but doesn't wait, so we have the same fundamental problem.
> 
> (I'd argue that the CLI ought to poll/loop until the process _actually_ dies before the `stop` command is considered complete, which would give the extension the behavior it wants, but there's a number of wait and timeout and PID reuse cases to consider. In any case, I've filed a [proposal](https://github.com/dapr/cli/issues/1160) for that.)
> 
> In any case, I'm going to close this PR for now. Thanks for the discussion!

Thanks for the response.

PR #265, opened 2023-01-20
Switch to `dapr list` to retrieve running applications

### LINKED ISSUE #264: Switch to `dapr list` to obtain running Dapr applications
Currently the extension periodically polls all processes to find running Dapr applications (filtering for those.  Originally this was necessary because the extension started `daprd` processes itself (as the only way then to start the sidecar independent of the application being debugged)and `dapr list` didn't yet exist or, even when it did, would only report processes started via `dapr`.  All of this is no longer necessary; the extension now (by default) uses `dapr` to start the sidecar and `dapr list` also is better at listing independent `daprd` instances.

The extension should now use `dapr list` to obtain running Dapr applications and dispense with process polling (and it's quirks across platforms).

philliphoff:
Instead of listing all processes and then filtering them down to just the Dapr processes, switches to using `dapr list` (and let the Dapr CLI worry about how to find them). This eliminates our dependency on `ps-list`, is simpler, and likely more performant.

Resolves #264


src/services/daprApplicationProvider.ts
```diff
@@ -134,12 +80,41 @@ export default class ProcessBasedDaprApplicationProvider extends vscode.Disposab
...
+                .withNamedArg('--output', 'json')
+                .build();
+
+        const result = await Process.exec(command);
+
+        if (result.code != 0) {
+            throw new Error(`'${command}' failed with exit code ${result.code}.`);
+        }
+
+        if (result.stderr.length > 0) {
```
bwateratmsft, line 95: This has bitten us in the past in Docker (assuming stderr output is an error), are you confident using it here?

bwateratmsft (approved):

src/services/daprApplicationProvider.ts
philliphoff: That's a good point; must just be muscle memory.  I've removed that check for now and we'll see how consistent Dapr is with its error codes.

PR #267, opened 2023-01-30
Add support for upcoming Dapr "run file" option

### LINKED ISSUE #202: Details pane is sluggish when changing between nodes

### LINKED ISSUE #203: Info in the Details pane does not clear when Tye is shutdown

philliphoff:
Adds support for the upcoming (v1.10) Dapr "run file" option in the `dapr` task.  Switches to `dapr version` from `dapr --version` (which is more deterministic).  Also fixes some latency in the "details" view (where it might take until the next app list refresh to update the view).

Mostly some prep work to adding further "run file" debugging integration.

Resolves #202
Resolves #203


src/services/daprCliClient.ts
```diff
@@ -20,6 +20,11 @@ export interface DaprCliClient {
     stopApp(application: DaprApplication | undefined): void;
 }
 
+interface DaprCliVersion {
+    'Cli version'?: string;
```
bwateratmsft, line 24: Ugh, is the CLI really spitting out property names with _spaces_?

bwateratmsft (approved):

src/services/daprCliClient.ts
philliphoff: Yes, I was surprised too.  It's a bit weird, but using `dapr version` instead of `dapr --version` gets around an issue where the actual value has random spaces appended to the actual value.

PR #287, opened 2023-02-16
Adds ability to browse to Dapr application entrypoint.

### LINKED ISSUE #275: [Feature] Open browser to Dapr application port
To save developer's time, enable them to open a browser to a running Dapr application's port.  If the application is part of a run file, and that run file indicates an endpoint to browse to (e.g. the Swagger endpoint), use that, too.

philliphoff:
For applications that expose a port, we add an inline context command to open a browser directly to that entrypoint, saving the developer the step of typing out the path in a browser.

> Note: there's no guarantee that the base endpoint is served by the application but, currently, there's no place for developers to indicate an appropriate subpath.  (That will hopefully be added to the Dapr runtime in due course.)

> Note: the extension also assume the application exposes an HTTP port.  (It might expose a gRPC port instead but, again, the Dapr runtime doesn't provide a way to distinguish between them.)

Screenshot 2023-02-16 at 11 25 42

Resolves #275

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,5 +1,5 @@
 {
-    "vscode-dapr.applications.browse.title": "Browse to Application",
+    "vscode-dapr.applications.browse.title": "Open Application in Browser",
     "vscode-dapr.applications.debug.title": "Debug Application",
     "vscode-dapr.applications.invoke-get.title": "Invoke (GET) Application Method",
     "vscode-dapr.applications.invoke-post.title": "Invoke (POST) Application Method",
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,4 +1,5 @@
 {
+    "vscode-dapr.applications.browse.title": "Browse to Application",
     "vscode-dapr.applications.debug.title": "Debug Application",
     "vscode-dapr.applications.invoke-get.title": "Invoke (GET) Application Method",
     "vscode-dapr.applications.invoke-post.title": "Invoke (POST) Application Method",
--- a/src/commands/applications/browseToApplication.ts
+++ b/src/commands/applications/browseToApplication.ts
@@ -20,7 +20,7 @@ async function browseToApplication(application: DaprApplication, ui: UserInput):
 }
 
 const createBrowseToApplicationCommand = (ui: UserInput) => (context: IActionContext, node: DaprApplicationNode | undefined): Promise<void> => {
-    if (node == undefined) {
+    if (node === undefined) {
         throw new Error(localize('commands.applications.browseToApplication.noPaletteSupport', 'Browsing requires selecting an application in the Dapr view.'));
     }
 
--- a/src/commands/applications/browseToApplication.ts
+++ b/src/commands/applications/browseToApplication.ts
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as nls from 'vscode-nls';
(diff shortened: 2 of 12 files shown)


package.nls.json
```diff
@@ -1,4 +1,5 @@
 {
+    "vscode-dapr.applications.browse.title": "Browse to Application",
```
bwateratmsft, line 2: Or maybe "Open Application in Browser"? At least to me it doesn't immediately click that this is going to open a browser window pointed at the app. It seems like it could also be opening something in the file explorer in VSCode.

src/commands/applications/browseToApplication.ts
```diff
@@ -0,0 +1,30 @@
...
+import { UserInput } from '../../services/userInput';
+
+const localize = nls.loadMessageBundle(getLocalizationPathForFile(__filename));
+
+async function browseToApplication(application: DaprApplication, ui: UserInput): Promise<void> {
+    if (application.appPort === undefined) {
+        throw new Error(localize('commands.applications.browseToApplication.noAppPort', 'No port is associated with the application \'{0}\'.', application.appId));
+    }
+
+    // TODO: Is this app always HTTP?
```
bwateratmsft, line 18: Docker uses the port as a hint. It's a rough heuristic but can help.

src/commands/applications/browseToApplication.ts
```diff
@@ -0,0 +1,30 @@
...
+    await ui.openExternal(`http://localhost:${application.appPort}`);
+}
+
+const createBrowseToApplicationCommand = (ui: UserInput) => (context: IActionContext, node: DaprApplicationNode | undefined): Promise<void> => {
+    if (node == undefined) {
```
bwateratmsft, line 23: I never remember what `==` actually does, so I'd prefer this:
```suggestion
    if (!!node) {
```
Then any falsy value goes into the `if` body.

bwateratmsft (approved):

src/commands/applications/browseToApplication.ts
philliphoff: We know the port is either HTTP or gRPC, we just don't know which.  The Dapr runtime does; I'm going to file an issue to have that property exposed.

bwateratmsft: Ah, I was thinking it was HTTP or HTTPS.

philliphoff: ...and it's *most likely* HTTP; my impression is that few Dapr applications expose a gRPC endpoint to the Dapr sidecar.

package.nls.json
philliphoff: I like that.  Updated.

src/commands/applications/browseToApplication.ts
philliphoff: That's a good catch!  I've been writing too much C# lately.  It should be `===` here and in a number of other, similar places (which I've now updated).  I've never liked the `!!value` pattern for some reason.

philliphoff: I should see if there's a linter warning I can enable for that; I don't every think I've legitimately used `==`.

bwateratmsft: I love the `!!syntax`! It's like yelling it but in reverse.

philliphoff: There is!  And I've turned it on (and caught another issue).

[eqeqeq ES Lint Rule](https://eslint.org/docs/latest/rules/eqeqeq)

PR #292, opened 2023-03-22
chore: enhance dapryml schema validation

IvanJobs:
This PR comes to a stage when we can discuss specific topics.

### DIFF
--- a/assets/schemas/dapr.io/dapr/cli/run-file.json
+++ b/assets/schemas/dapr.io/dapr/cli/run-file.json
@@ -27,7 +27,8 @@
                             "appPort": {
                                 "description": "The port your application is listening on.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "command": {
                                 "description": "The command and arguments used to start your application.",
@@ -39,27 +40,32 @@
                             "daprGRPCPort": {
                                 "description": "The gRPC port for Dapr to listen on.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "daprHTTPPort": {
                                 "description": "The HTTP port for Dapr to listen on.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "daprInternalGRPCPort": {
                                 "description": "The gRPC port for the Dapr Internal API to listen on.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "metricsPort": {
                                 "description": "The port for the metrics server.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "profilePort": {
                                 "description": "The port for the profile server.",
                                 "type": "integer",
-                                "minimum": 0
+                                "minimum": 0,
+                                "maximum": 65535
                             },
                             "unixDomainSocket": {
                                 "description": "Path to a unix domain socket dir. If specified, Dapr API servers will use Unix Domain Sockets.",
--- a/assets/schemas/dapr.io/dapr/cli/run-file.json
+++ b/assets/schemas/dapr.io/dapr/cli/run-file.json
@@ -139,7 +139,7 @@
                     "type": "integer",
                     "minimum": 0
                 },
-                "daprPath": {
+                "runtimePath": {
                     "description": "The path to the Dapr runtime (i.e. daprd).",
                     "type": "string"
                 },
--- a/assets/schemas/dapr.io/dapr/cli/run-file.json
+++ b/assets/schemas/dapr.io/dapr/cli/run-file.json
@@ -21,7 +21,8 @@
                             },
                             "appID": {
                                 "description": "An ID for your application, used for service discovery.",
-                                "type": "string"
+                                "type": "string",
+                                "pattern": "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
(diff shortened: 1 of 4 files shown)


assets/schemas/dapr.io/dapr/cli/run-file.json
```diff
@@ -21,7 +21,8 @@
                             },
                             "appID": {
                                 "description": "An ID for your application, used for service discovery.",
-                                "type": "string"
+                                "type": "string",
+                                "pattern": "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
```
IvanJobs, line 25: I didn't found any official regex for validating app id. This regex was found on stackoverflow.

assets/schemas/dapr.io/dapr/cli/run-file.json
```diff
@@ -156,7 +157,10 @@
                 },
                 "env": {
                     "description": "Environment variables to be passed to the application.",
-                    "type": "string"
+                    "type": "object",
```
IvanJobs, line 160: `env` is not in string type (I believe it's incorrect). 
`env` should be a map or object, so change the type to `object`.

assets/schemas/dapr.io/dapr/cli/run-file.json
```diff
@@ -156,7 +157,10 @@
...
+                    "additionalProperties": {
+                        "type": ["string", "number", "boolean"]
```
IvanJobs, line 162: Here I add "number" and "boolean" here for compatibility so that developers don't have to add quotes for cases like below:
```
env:
   SOME_ENV1: true
   SOME_ENV2: 123
```
I'm not very sure about this change actually. We may discuss about it.

assets/schemas/dapr.io/dapr/cli/run-file.json
```diff
@@ -165,7 +169,8 @@
                 },
                 "placementHostAddress": {
                     "description": "The host on which the placement service resides.",
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$"
```
IvanJobs, line 173: This one is similiar with the previous. I tried my best to find an regex here.

package.json
```diff
@@ -12,6 +12,9 @@
 		"workspace",
 		"ui"
 	],
+	"extensionDependencies": [
+		"redhat.vscode-yaml"
```
IvanJobs, line 16: YAML file's intelli-sense depends on this extension.

assets/schemas/dapr.io/dapr/cli/run-file.json
```diff
@@ -138,7 +145,7 @@
                     "type": "integer",
                     "minimum": 0
                 },
-                "daprPath": {
+                "runtimePath": {
```
yukun-dong, line 148: the "daprPath" has been changed to "runtimePath" in dapr/cli (https://github.com/dapr/cli/pull/1203#issuecomment-1416229659), so here should change as well

philliphoff: Good catch!

package.json
philliphoff: Because IntelliSense isn't a critical feature of the extension--helpful, sure, but not absolutely needed--I had intended *not* to have a hard dependency on the YAML extension.  Unfortunately, there's no way for extensions themselves to just *recommend* extensions, which I believe would be a better experience.  (VS Code itself has a mechanism for arbitrary recommendations, but it is something we would have to ask for and I don't know if there's a strong enough argument for it.)

If we wanted, we could hook the opening `dapr.yaml` files and then look to see whether the YAML extension is installed; if not, then we could display our own recommendation.

assets/schemas/dapr.io/dapr/cli/run-file.json
philliphoff: For "standalone" (i.e. local) mode, it [looks](https://github.dev/dapr/dapr/blob/a66a3357fda7898c1ee34df6492b311404f2d591/pkg/validation/validation.go#L45) like `daprd` only disallows the use of periods (`.`) in app IDs (which is different than in Kubernetes mode, where the app ID must conform to the Kubernetes label format.  I might use a pattern that only excludes periods.

assets/schemas/dapr.io/dapr/cli/run-file.json
philliphoff: I think that's fine, as long as those variations work as expected when the applications are run.

assets/schemas/dapr.io/dapr/cli/run-file.json
philliphoff: The placement host can include an optional port value (e.g. can be `localhost` or `localhost:5555`).  Using regular expressions for (even portions of) URLs can be tricky, especially if we have to consider both name and IP (v4/v6) address forms.  We might avoid doing so for now.

philliphoff:
While this is a good start, there's potentially a higher-level set of validation that could be useful to developers.  For example, where a property requires a path, ensure that that path exists or, when the developer is adding the port, provide a default that matches the application settings.  That's more semantic analysis rather than the syntactic done through this JSON schema and is done through [language servers and related APIs](https://code.visualstudio.com/api/language-extensions/overview).

assets/schemas/dapr.io/dapr/cli/run-file.json
IvanJobs: It should be fine since dapr cli used yaml.unMarshal to deserialize the dapr.yaml file and it's compatible for boolean & int for string types.
https://github.com/dapr/cli/blob/4e3ae81df53a87c4707e95f5c5dfd8295e230c80/pkg/standalone/runfileconfig/run_file_config_parser.go#L36

philliphoff (approved):

PR #297, opened 2023-06-02
feat: task use existing run file

yukun-dong:
https://github.com/microsoft/vscode-dapr/issues/272

### DIFF
--- a/src/commands/scaffoldDaprTasks.ts
+++ b/src/commands/scaffoldDaprTasks.ts
@@ -187,8 +187,8 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
     };
 
     if (fs.existsSync(runFilePath)) {
-        const runFileTask: vscode.MessageItem = { title: 'yes' };
-        const defaultTask: vscode.MessageItem = { title: 'no' };
+        const runFileTask: vscode.MessageItem = { title: localize('commands.scaffoldDaprTasks.useExistingDaprRunFile', 'yes') };
+        const defaultTask: vscode.MessageItem = { title: localize('commands.scaffoldDaprTasks.notUseExistingDaprRunFile', 'no') };
         const result = await ui.showWarningMessage('You already have a Dapr run file. Would you like to use it in the scaffolded task?',
             { modal: true },
             runFileTask, defaultTask);
@@ -198,7 +198,7 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
                 'dapr', folder,
                 label => {
                     const daprUpTask: DaprTaskDefinition = {
-                        label: label,
+                        label,
                         type: 'dapr',
                         runFile: '${workspaceFolder}/dapr.yaml',
                     };
@@ -209,11 +209,11 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
                 'Launch Dapr', folder,
                 name => {
                     const daprDebugConfiguration: DaprDebugConfiguration = {
-                        "name": name,
-                        "request": "launch",
-                        "type": "dapr",
-                        "runFile": "${workspaceFolder}/dapr.yaml",
-                        "preLaunchTask": preLaunchTask
+                        name,
+                        request: 'launch',
+                        type: 'dapr',
+                        runFile: '${workspaceFolder}/dapr.yaml',
+                        preLaunchTask
                     };
                     return daprDebugConfiguration;
                 }, onConfigConflict);
--- a/src/commands/scaffoldDaprTasks.ts
+++ b/src/commands/scaffoldDaprTasks.ts
@@ -15,7 +15,6 @@ import { ConflictHandler, ConflictUniquenessPredicate } from '../scaffolding/con
 import { names, range } from '../util/generators';
 import { getLocalizationPathForFile } from '../util/localization';
 import { TextDecoder } from 'util';
-import * as fs from 'fs';
 import path from 'path';
 import { DaprDebugConfiguration } from '../debug/daprDebugConfigurationProvider';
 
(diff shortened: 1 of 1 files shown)


src/commands/scaffoldDaprTasks.ts
```diff
@@ -134,6 +138,89 @@ async function createUniqueName(prefix: string, isUnique: ConflictUniquenessPred
...
+            return { 'type': 'overwrite' };
+        } else {
+            name = await createUniqueName(localize('commands.scaffoldDaprTasks.configurationPrefix', '{0} - ', name), isUnique);
+
+            return { 'type': 'rename', name };
+        }
+    };
+
+    if (fs.existsSync(runFilePath)) {
+        const runFileTask: vscode.MessageItem = { title: 'yes' };
```
philliphoff, line 190: Nit: these strings should be localized.

src/commands/scaffoldDaprTasks.ts
```diff
@@ -134,6 +138,89 @@ async function createUniqueName(prefix: string, isUnique: ConflictUniquenessPred
...
+        const defaultTask: vscode.MessageItem = { title: 'no' };
+        const result = await ui.showWarningMessage('You already have a Dapr run file. Would you like to use it in the scaffolded task?',
+            { modal: true },
+            runFileTask, defaultTask);
+
+        if (result === runFileTask) {
```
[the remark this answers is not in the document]
yukun-dong: In my opinion it's ok to have 2 sets of defining tasks and configurations. Below are the reasons:

1. Although they are all dapr related tasks and configurations, the properties in them are quite different. 
2. The flow to generate them is different: dapr run file doesn't require additional user input, not using dapr run file requires user input of appId, appPort, etc.

Therefore, I don't think it's convenient to define these task and configuration in one place.

Please let me know if you have additional comments on this. Thank you!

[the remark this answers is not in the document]
philliphoff: In that case, then we should use constants that are shared between them as much as possible, to ensure that values stay the same between versions.

src/commands/scaffoldDaprTasks.ts
```diff
@@ -242,73 +167,199 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
...
+        if (result === overwrite) {
+            return { 'type': 'overwrite' };
+        } else {
+            name = await createUniqueName(localize('commands.scaffoldDaprTasks.configurationPrefix', '{0} - ', name), isUnique);
+
+            return { 'type': 'rename', name };
+        }
+    };
+
+    if (fs.existsSync(runFilePath)) {
```
philliphoff, line 190: Nit: when using `fs` we should use the async versions (in `fs/promises`) when possible. That said, we should try to use VS Code's [filesystem API](https://code.visualstudio.com/api/references/vscode-api#FileSystem) even before that, as that helps enable the extension to run in virtualized environments (like browsers). (There are still some references to `fs/promises` in the extension that we should replace when convenient, but ideally we should not create new ones.)

src/commands/scaffoldDaprTasks.ts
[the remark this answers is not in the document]
yukun-dong: The 2 sets of task and configuration are listed below. The only shared constants are type: "dapr" and others are quite different. So I used constants for "dapr" and others remain the same. I wonder if that's ok.

Not using runFile:
```json
        {
            "appId": "app",
            "appPort": 5074,
            "label": "dapr-debug",
            "type": "dapr",
            "dependsOn": "build"
        },
        {
            "appId": "app",
            "label": "daprd-down",
            "type": "daprd-down"
        },
```
```json
        {
            "name": ".NET Core Launch (web) with Dapr",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "dapr-debug",
            "program": "${workspaceFolder}/bin/Debug/net7.0/netcoreapp.dll",
            "args": [],
            "cwd": "${workspaceFolder}",
            "stopAtEntry": false,
            "serverReadyAction": {
                "action": "openExternally",
                "pattern": "\\bNow listening on:\\s+(https?://\\S+)"
            },
            "env": {
                "ASPNETCORE_ENVIRONMENT": "Development"
            },
            "sourceFileMap": {
                "/Views": "${workspaceFolder}/Views"
            },
            "postDebugTask": "daprd-down"
        },
```
Using runFile:
```json
        {
            "label": "dapr",
            "type": "dapr",
            "runFile": "$(workspaceFolder)/dapr.yaml"
        }
```
```json
        {
            "name": "Launch Dapr",
            "request": "launch",
            "type": "dapr",
            "runFile": "$(workspaceFolder)/dapr.yaml",
            "preLaunchTask": "dapr"
        }
```

src/commands/scaffoldDaprTasks.ts
yukun-dong: Thanks! I updated it to use `vscode.workspace.fs.stat()` method.

src/commands/scaffoldDaprTasks.ts
```diff
@@ -242,73 +171,199 @@ export async function scaffoldDaprTasks(context: IActionContext, scaffolder: Sca
...
+            return { 'type': 'overwrite' };
+        } else {
+            name = await createUniqueName(localize('commands.scaffoldDaprTasks.configurationPrefix', '{0} - ', name), isUnique);
+
+            return { 'type': 'rename', name };
+        }
+    };
+
+    if (await vscode.workspace.fs.stat(vscode.Uri.file(runFilePath)).then(() => true, () => false)) {
+        const runFileTask: vscode.MessageItem = { title: localize('commands.scaffoldDaprTasks.useExistingDaprRunFile', 'yes') };
```
philliphoff, line 195: Rather than `yes` and `no`, I think I'd go with something like `Use Run File` and `Run Dapr Directly`.

philliphoff (approved):
Just one comment re: text in user prompts but otherwise looks ready to go.

PR #298, opened 2023-06-09
Updates to dependencies

philliphoff:
Updates dependencies, some of which (i.e. `glob`) required some minor reworkd due to breaking changes.


package.json
```diff
@@ -735,48 +735,48 @@
...
-		"webpack": "^5.76.0",
-		"webpack-cli": "^4.9.2"
+		"typescript": "^5.1.3",
+		"vsce": "^2.15.0",
+		"vscode-nls-dev": "^4.0.4",
+		"webpack": "^5.86.0",
+		"webpack-cli": "^5.1.4"
 	},
 	"dependencies": {
 		"@microsoft/vscode-azext-utils": "^0.3.7",
```
bwateratmsft, line 765: Long as we're updating all the things, I'd recommend bumping this one to `^2.0.0`, it will remove a _lot_ of dependencies that might (will) bring CG alerts in the future

bwateratmsft (approved):

package.json
philliphoff: I'll take a look; I was worried about back-compat, as I recall Dapr relied on some APIs that got heavily refactored at some point and not as easily accessible.

philliphoff: No immediate build or other noticeable issues, so I'll go with that for now.

Issue #284, opened 2023-02-10
[Enhancement] Infer `dapr.yaml` when omitted from configuration


philliphoff:
If a Dapr task definition or debug launch configuration has been defined without an explicit run file, infer the `dapr.yaml` at the root of the workspace (if present).  This will make creation/scaffolding of such configuration files simpler.

IvanJobs:
@MregXN You can investigate on this one first.

MregXN:
Hi, @philliphoff. Is that mean the local dapr.yaml file should be read by this extension and properties in this file needs to be filled in task.json and launch.json?

philliphoff:
No, the suggestion is that, if the `tasks.json` and `launch.json` files omit the property that would normally point to the one and only `dapr.yaml` in the repo, the extension, when running the task and/or launching the debugger would simply assume that that `dapr.yaml` is the one to be used in the `dapr run -f ...` command.  The `.json` files need not be changed.  We're just saying that the user doesn't need to tell the extension (in the `.json` files) which file to use because the extension can figure it out *at the time it actually needs to use it*.

MregXN:
Hi, @philliphoff. I wanna confirm my understanding is correct with following code snippet. 

tasks.json:
```json
      {
          "label": "dapr",
          "type": "dapr",
         // "runFile": "./dapr.yaml"
      }
```
launch.json:
```json
      {
          "name": "Launch Dapr",
          "request": "launch",
          "type": "dapr",
          //"runFile": "${workspaceFolder}/dapr.yaml",
          "preLaunchTask": "dapr"
      }
```

If users do not specify a runFile (e.g., it is commented out in the code snippet), but task and launch configurations of type "dapr" still exist, extension should automatically use the dapr.yaml in the workspace without modifying these two json files. Is this right?

philliphoff:
Closed with #300.

PR #300, opened 2023-06-14
Infer `dapr.yaml` when omitted from configuration

MregXN:
Please reference the issue: #284


src/tasks/daprCommandTaskProvider.ts
```diff
@@ -79,12 +79,18 @@ export default class DaprCommandTaskProvider extends CommandTaskProvider {
                                 .withNamedArg('--unix-domain-socket', daprDefinition.unixDomainSocket)
                                 .withArgs(daprDefinition.args)
                                 .withArgs(daprDefinition.command)
-                                .build();
+
                         
+                        // infer dapr.yaml when omitted from configuration
+                        if (daprDefinition.appId === undefined && daprDefinition.runFile === undefined) {
```
philliphoff, line 85: I wonder if there's any other property that should be checked to determine whether inferrence is warranted.  I *think* there's a way to use TS to get the list of properties defined on the Dapr task configuration interface, then loop over them looking for set values for those properties.  Only if none are set then would we infer the Dapr run file.

src/tasks/daprCommandTaskProvider.ts
```diff
@@ -79,12 +79,18 @@ export default class DaprCommandTaskProvider extends CommandTaskProvider {
...
+                            commandLineBuilder.withNamedArg('--run-file', "./dapr.yaml")
```
philliphoff, line 86: Should we first validate that the Dapr run file exists?

MregXN: There may be a premise that needs to be confirmed again: if the configurations are omitted by the user, should this extension run "dapr.yaml" automatically even if users may use a different name for this YAML file?

If the answer is YES, I think there is no need to validate the run file is exists because the command "dapr run -f ..." will do that and return ERROR in vscode terminal.

src/tasks/daprCommandTaskProvider.ts
MregXN: modified in latest commit.

src/tasks/daprCommandTaskProvider.ts
```diff
@@ -79,12 +87,39 @@ export default class DaprCommandTaskProvider extends CommandTaskProvider {
...
-                        
-                        return callback(command, { cwd: definition.cwd });
+
+
+                        // infer dapr.yaml when omitted from configuration
+                        for (const def in daprDefinition) {
+                            // skip daprDefinition.type as it will be set by default
+                            // if any property being set means do not need to infer dapr.yaml
+                            // eslint-disable-next-line no-prototype-builtins
+                            if (daprDefinition.hasOwnProperty(def) && def !== "type" && def !== undefined) {
```
philliphoff, line 97: There may be other properties that we want to exclude, e.g. `label`, `cwd`, `dependsOn`, etc.  Perhaps we should have a hardcoded list?

src/tasks/daprCommandTaskProvider.ts
```diff
@@ -79,12 +87,39 @@ export default class DaprCommandTaskProvider extends CommandTaskProvider {
...
+
+                        if (!folder) {
+                            context.errorHandling.suppressReportIssue = true;
+                            throw new Error(localize('tasks.daprCommandTaskProvider.noFolderOrWorkspace', 'Open a folder or workspace.'));
+                        }
+            
+                        const runFilePath = path.join(folder.uri.fsPath, 'dapr.yaml');
+                        if (fs.existsSync(runFilePath)) {
+                            commandLineBuilder.withNamedArg('--run-file', runFilePath);
+                            const command = commandLineBuilder.build();
```
philliphoff, line 114: We should reuse the same command line builder as above, so that there's no risk of command lines diverging as the CLI changes.

src/tasks/daprCommandTaskProvider.ts
```diff
@@ -79,12 +87,39 @@ export default class DaprCommandTaskProvider extends CommandTaskProvider {
...
+                        // when all properties are undefined the command "dapr run --run-file ./dapr.yaml" will be excuted.
+                        const folder = vscode.workspace.workspaceFolders?.[0];
+
+                        if (!folder) {
+                            context.errorHandling.suppressReportIssue = true;
+                            throw new Error(localize('tasks.daprCommandTaskProvider.noFolderOrWorkspace', 'Open a folder or workspace.'));
+                        }
+            
+                        const runFilePath = path.join(folder.uri.fsPath, 'dapr.yaml');
+                        if (fs.existsSync(runFilePath)) {
```
philliphoff, line 112: Nit: when using `fs` we should use the async versions (in `fs/promises`) when possible.  That said, we should try to use VS Code's filesystem API even before that, as that helps enable the extension to run in virtualized environments (like browsers).  (There are still some references to `fs/promises` in the extension that we should replace when convenient, but ideally we should not create new ones.)

src/tasks/daprCommandTaskProvider.ts
MregXN: method `hasOwnProperty` will exclud these properties because they are inherited from `TaskDefinition` . It is not the own properties of DaprTaskDefinition.

src/tasks/daprCommandTaskProvider.ts
MregXN: The commandLineBuilder here is reused. In this file commandLineBuilder is initialized only once.

src/tasks/daprCommandTaskProvider.ts
MregXN: modify it in latest commit.

src/tasks/daprCommandTaskProvider.ts
philliphoff: Yes, sorry, you're right. What I *think* I was getting at was trying to avoid having to specify the `--run-file` argument twice (as that becomes a point of risk if/when the CLI arguments ever change).  It might work to capture the command line building into a separate function being passed a configuration object that can be set appropriately in each circumstance.

MregXN: add it in the latest commit

src/tasks/daprCommandTaskProvider.ts
```diff
@@ -44,57 +49,102 @@ export interface DaprTaskDefinition extends TaskDefinition {
...
+                        if (!folder) {
+                            context.errorHandling.suppressReportIssue = true;
+                            throw new Error(localize('tasks.daprCommandTaskProvider.noFolderOrWorkspace', 'Open a folder or workspace.'));
+                        }
+
+                        const runFilePath = path.join(folder.uri.fsPath, 'dapr.yaml');
+
+                        await checkFileExists(runFilePath).then((fileExists) => {
+                            if (fileExists) {
+                                const command = createCommandLineBuilder(daprPathProvider).withNamedArg('--run-file', runFilePath).build();
```
philliphoff, line 88: This still repeats the `--run-file` argument; what about:

`createCommandLineBuilder(daprPathProvider, { runFile: runFilePath).build();`

MregXN: modified in the latest commit

philliphoff (approved):

PR #322, opened 2024-03-25
Add 1ES pipeline for testing.

philliphoff:


.azure-pipelines/common/setup.yml
bwateratmsft, line 1: This file is no longer used. Instead, you'll need to create a `.nvmrc` file at the repo root.

philliphoff: I plan to refactor the former build system once the 1ES build system is in place and running, likely borrowing from the other extensions.

bwateratmsft: Ok. You'll still want an .nvmrc file.

philliphoff: Added.

bwateratmsft (approved):

Issue #319, opened 2024-01-11
Can't debug multi-run dapr solution (The task dapr cannot be tracked)


jasonshave:
# Overview:

I'm able to run my dapr solution which has 3 ASP.NET 8 web API's using the `dapr -f` multi-run file but having some difficulty getting the debug experience to work.

## Actual behavior:

When using the CTRL+SHIFT+P to scaffold a launch.json and task for tasks.json, it creates the following:

```json
{
  "name": "Launch Dapr",
  "request": "launch",
  "type": "dapr",
  "runFile": "${workspaceFolder}/dapr.yaml",
  "preLaunchTask": "dapr"
}
```

```json
{
  "label": "dapr",
  "type": "dapr",
  "runFile": "$(workspaceFolder)/dapr.yaml"
},
```

1. The first issue is that the runFile's value results in a file not found. If I remove the $(workspaceFolder) and just use dapr.yaml, the project starts up with the right components and sidecar/etc.
2. Once the solution is started, I get this error:

![image](https://github.com/microsoft/vscode-dapr/assets/28457536/61f2177c-ac4a-4d2a-8348-83eb3c8971f3)

If I click DebugAnyway, I get:

![image](https://github.com/microsoft/vscode-dapr/assets/28457536/e2a00e27-9abb-4727-ad20-ca24c749318f)

3. If I use the dapr extension to locate the running apps and click on any of the debug buttons I get:

![image](https://github.com/microsoft/vscode-dapr/assets/28457536/a0528748-1f68-491e-a6ba-99c4bc396b93)

## Expected behavior:

I should be able to attach and debug with breakpoints in each independent application.

joshfreitas1984:
First up your tasks are wrong you can just do:
{
    "label": "dapr",
    "type": "dapr",
    "runFile": "./dapr.yaml"
}

Was getting the same error about fastlist so I downloaded it chucked it in the \vendor but then it never finds the process - I'm guessing somethings busted. 

Is this extension abandoned?

karpikpl:
same issue happening for me.
that `"runFile": "$(workspaceFolder)/dapr.yaml"` is actually how DAPR extension scaffolds the project

joshfreitas1984:
Anyone that finds this - pretty sure the plugin is abandoned. However, .NET Aspire is your friend! The f5 experience on the Apphost is amazing it will load up all your hosts and attach the debuggers super fast.

philliphoff:
[...]

PR #329, opened 2024-05-30
Resolve component scaffolding and configuration issues

philliphoff:
Resolves a couple of long-standing issues:

 - No longer scaffolds a (now deprecated) Zipkin component (see: [dapr/dapr#2960](https://github.com/dapr/dapr/issues/2960))
 - Allows `resourcesPath` and `resourcesPaths` properties be used for `dapr` tasks (instead of only `daprd` tasks)
 - Allows use of variable substitution (e.g. `${workspaceFolder}`) in Dapr-related task definitions and fixes a scaffolding error (which used `(` instead of `{`) (see: #319)


src/tasks/daprCommandTaskProvider.ts
```diff
@@ -134,6 +134,8 @@ function createCommandLineBuilder(daprPathProvider: () => string, daprDefinition
             .withNamedArg('--metrics-port', daprDefinition.metricsPort)
             .withNamedArg('--placement-host-address', daprDefinition.placementHostAddress)
             .withNamedArg('--profile-port', daprDefinition.profilePort)
+            .withNamedArg('--resources-path', daprDefinition.resourcesPath)
+            .withArrayArgs('--resources-path', daprDefinition.resourcesPaths)
```
bwateratmsft, line 138: Does this need to be like this:
```suggestion
            .withArrayArgs('--resources-paths', daprDefinition.resourcesPaths)
```
I'm guessing no but I want to double check there wasn't a copy-paste error here.

bwateratmsft (approved):

src/tasks/daprCommandTaskProvider.ts
philliphoff: Correct (i.e. no).  It's copied from the `daprd` task.  The first uses the now-deprecated `resourcesPath` string property (to specify only a single value) while the second uses the new `resourcesPaths` string array property to allow more than one specified path--the CLI allows it to be specified multiple times.

